PR #52, opened 2021-11-04
VIDEO-7631: Use one sync service per stream

timrozum:
### JIRA link(s):

- [VIDEO-7631](https://issues.corp.twilio.com/browse/VIDEO-7631)

### Description

1. Create a `Live Video Backend Storage` sync service during deploy to store state that is only used by the backend. It has a `streams` map with room SID for key and a few other SIDs are stored in the map item data. Eventually the sync webhook may need to know room SID and I think there are some ways we could do that.
2. Create a sync service for each stream when the stream is created. This prepares us to use sync reachability which is scoped to a sync service. And it simplifies sync object names because we don't have to include room SID anymore.
3. No client changes were required so both web and iOS apps work.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/serverless/functions/create-stream.js
+++ b/serverless/functions/create-stream.js
@@ -14,7 +14,8 @@ module.exports.handler = async (context, event, callback) => {
     TWILIO_API_KEY_SID,
     TWILIO_API_KEY_SECRET,
     CONVERSATIONS_SERVICE_SID,
-    SYNC_SERVICE_SID,
+    BACKEND_STORAGE_SYNC_SERVICE_SID,
+    SYNC_SERVICE_NAME_PREFIX,
     DOMAIN_NAME,
   } = context;
 
@@ -48,10 +49,9 @@ module.exports.handler = async (context, event, callback) => {
     return callback(null, response);
   }
 
-  let room, playerStreamer, mediaProcessor, streamDocument, conversation, raisedHandsMap;
+  let room, playerStreamer, mediaProcessor, streamSyncService, streamSyncClient, conversation, raisedHandsMap;
 
   const client = context.getTwilioClient();
-  const syncClient = client.sync.services(context.SYNC_SERVICE_SID);
 
   try {
     room = await client.video.rooms.create({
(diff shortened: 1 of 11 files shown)


serverless/functions/create-stream.js
```diff
@@ -111,32 +111,54 @@ module.exports.handler = async (context, event, callback) => {
...
       error: {
-        message: 'error creating stream document',
+        message: 'error adding stream to streams map',
         explanation: e.message,
       },
     });
     return callback(null, response);
   }
 
-  const raisedHandsMapName = `raised_hands-${room.sid}`
```
aaalaniz, line 135: Yes! Awesome

serverless/functions/join-stream-as-speaker.js
```diff
@@ -57,9 +56,26 @@ module.exports.handler = async (context, event, callback) => {
...
+  } catch (e) {
+    response.setStatusCode(500);
+    response.setBody({
+      error: {
+        message: 'error finding stream map item',
+        explanation: e.message,
+      },
+    });
+    return callback(null, response);
+  }
```
aaalaniz, line 72: Might be worth adding this to the common code

https://github.com/twilio/twilio-live-interactive-video/blob/main/serverless/middleware/common.private.js

aaalaniz (changes requested):
Nice work!

serverless/functions/join-stream-as-speaker.js
timrozum: I pushed some changes for this. I think it's a small step in the right direction. It does reduce duplicate code a little. What I don't like about it is:

1. I don't have a way to reuse the error handling which is the bulk of the code. I think the large amount of error handling boilerplate makes it difficult to follow the backend code.
2. Having all reusable code in a single file does not seem like it will scale well. I would prefer to have something like `backend-storage.private.js` for anything related to backend storage. And ideally it completely abstracts the fact that it is backed by sync.

I have ideas but I think I need to team up with a JS pro before going down this path more so probably best to wait until Tim M is back.

aaalaniz: Fair enough. Thanks for reducing some of the duplicate code!

aaalaniz (approved):

PR #53, opened 2021-11-04
VIDEO-7597: Enable sync reachability

timrozum:
### JIRA link(s):

- [VIDEO-7597](https://issues.corp.twilio.com/browse/VIDEO-7597)

### Description

1. Enable reachability on the stream sync service so we can handle user connection status better. Reachability debouncing is enabled so that we don't get a disconnect event when connections are rebalanced hourly. I think eventually we may want to add more logic in the webhook so that when we get a disconnect event we take extra steps to force the user to be disconnected. This is because I think there could be inconsistent state if the user reconnected after the debounce window. In this situation it's probably better to just disconnect them because this will not be a good state to be in.
2. Make webhook remove user from the raised hands map when a disconnect event is received.
3. Make changes to the backend so that the viewer document is created for both viewers and speakers. This simplifies user role changes on the client. I also want to rename this file from `viewer` to `user` because there really is no reason not too and it is a more general purpose way to push state to a specific user. I did not make this change in this PR because it would break the web app. 
4. Make changes to the iOS app so that sync client and objects are not torn down when changing role. The sync client in particular must stay connected or the webhook will receive a disconnect event. I think the web app already behaves this way but we aren't yet doing anything inside the webhook that would cause an issue in any case.

We can and will do more with the sync webhook but initially I just wanted to enable reachability and use it for one thing (remove the user from the raised hands map).

To test I raised my hand on one device and then force quit the app. A few seconds later the raised hand went away on the other device which is very cool! 😎

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


serverless/functions/join-stream-as-speaker.js
```diff
@@ -72,25 +72,44 @@ module.exports.handler = async (context, event, callback) => {
   }
 
   const streamSyncClient = client.sync.services(streamMapItem.data.sync_service_sid);
-  
+
+  const viewerDocumentName = `viewer-${user_identity}`;
+  // Create viewer document
   try {
-    // Reset the viewers hand_raised and speaker_invite status
-    await streamSyncClient.documents(`viewer-${user_identity}`).update({ data: { speaker_invite: false } });
```
timrozum, line 78: We were updating the viewer document to clear the speaker invite but I don't think we needed to. This should only be important when the user is a viewer and that happens in `join-stream-as-viewer`.

apps/ios/LiveVideo/LiveVideo/Twilio/Sync/SyncManager.swift
```diff
@@ -8,6 +8,7 @@ import TwilioSyncClient
 /// Consolidates configuration and error handling for stores that are backed by [Twilio Sync](https://www.twilio.com/sync).
 class SyncManager: NSObject {
     let errorPublisher = PassthroughSubject<Error, Never>()
+    var isConnected: Bool { client != nil }
```
aaalaniz, line 11: Does the Sync client provide a state that provides a more accurate representation of the connected state?

serverless/functions/sync-webhook.protected.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+exports.handler = async function (context, event, callback) {
+  const common = require(Runtime.getAssets()['/common.js'].path);
+  const { response } = common(context, event, callback);
+  const client = context.getTwilioClient();
+
+  if (event.EventType === 'endpoint_disconnected') {
```
aaalaniz, line 8: Nice 👍🏼 

What other cases will this webhook cover?

* Removing a viewer for speakers and viewers

Any others?

aaalaniz (changes requested):

apps/ios/LiveVideo/LiveVideo/Twilio/Sync/SyncManager.swift
timrozum: There isn't a simple way to check if the client is not completely dead.

I could probably do:
```swift
var isConnected: Bool {
    guard let client = client else { return false }

    switch client.connectionState {
    case .unknown, .connected, .connecting, .error:
        return true
    case .fatalError, .denied, .disconnected:
        return false
    @unknown default:
        return true
    }
}
```
The `error` state is not fatal but this isn't super clear from API or docs.

And `@unknown default:` is because of Objc bridging. I guess I assume any new state that is added is connected.

And then there is the connection rebalancing that happens hourly that may impact this. It shouldn't go into `disconnected` but sort of hard to test. 

So I tend to think relying on `connectionState` as little as possible is good given everything going on and probably best to just check for nil client. If we receive a fatal error the client is set to nil. 

I could be convinced otherwise though. I understand the intent of making it safer. I'm just not sure the added complexity actually makes it safer.

serverless/functions/sync-webhook.protected.js
timrozum: Yes removing viewers and speakers from sync. Probably cleaning up sync state is the main thing.

I think we probably want to keep using the room webhook for other clean up when the room is empty. I think the room has a 30 second connection timeout. Using sync for this would be more immediate but I think the room webhook is probably more reliable and simpler.

That might be it.

apps/ios/LiveVideo/LiveVideo/Twilio/Sync/SyncManager.swift
aaalaniz: Okay fair enough. I was mainly trying to avoid cases where sync is in a terminal failed state and we proceed with connecting to a room or player.

aaalaniz (approved):

PR #54, opened 2021-11-09
VIDEO-7665 End active player streamers and media processors when removing the app

aaalaniz:
**Contributing to Twilio**

### JIRA link(s):

- [VIDEO-7665](https://issues.corp.twilio.com/browse/VIDEO-7665)

### Description

The app relies on webhooks to end Twilio Live resources. As a result, when a user removes the app from Twilio Runtime, remove all media processors and player streamers. This action prevents streams from continuing after an app has been explicitly removed by a developer.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


serverless/scripts/remove.js
```diff
@@ -29,9 +29,35 @@ async function remove() {
 
   cli.action.start('Removing Sync Services');
   const syncServices = await client.sync.services.list();
-  syncServices.forEach(function(service) {
+  syncServices.forEach(async function(service) {
     if (service.friendlyName.includes(constants.SYNC_SERVICE_NAME_PREFIX)) {
-      client.sync.services(service.sid).remove();
+      cli.action.start(`Removing Sync Service: ${service.friendlyName}`);
+      const syncService = client.sync.services(service.sid);
```
timrozum, line 35: Should this be called `syncClient`? Note that the `forEach` parameter is already a "service". I think this is the convention @timmydoza started.

timrozum (changes requested):
This is cool!

serverless/scripts/remove.js
aaalaniz: I think `syncService` is the correct descriptor here. The `syncService` is a `SyncContext` object according to the API docs so I think the operations match the descriptor. However, I can be more explicit and call it a `liveSyncService` which is more accurate. thoughts?

timrozum: I don't think we need to prepend with `live`. 

What doc are you referring to?

If I look at the helper library source I see `ServiceInstance` and `ServiceContext`. Interesting that we aren't using either of those terms.

What is confusing to me is how we use `service` to create `syncService`. Which I assume is why Tim M used the convention `syncClient`. To distinguish between an instance and context. We shouldn't call both an instance and a context a service right? 

What would we call these things here?
https://github.com/twilio/twilio-live-interactive-video/blob/4bb4e68d3fd7afdea91f1c341027312d2fd68d5c/serverless/functions/create-stream.js#L116-L123

I'm inclined to use the established convention until Tim M returns or there is an obvious better way. 

But I won't hold this up since it is just semantics and I am no JS pro.

timrozum (approved):

PR #55, opened 2021-11-09
Add viewers to participants screen

timrozum:
### JIRA link(s):

- [VIDEO-7599](https://issues.corp.twilio.com/browse/VIDEO-7599)

### Description

1. Create new sync map to store the set of viewers so that clients can display who is viewing the stream. 
2. Update the iOS participants screen to show who is viewing the stream (not just who has their hand raised).
3. Created new `viewer-connected-to-player` endpoint so that clients can inform the backend when a user is connected to the player and viewing the stream. There is currently not a webhook to use for this so the simplest and most reliable solution was to create this new endpoint.
4. This PR will not break the web app but the web app won't display viewers or inform the backend when a viewer has connected to the player.
5. During testing everything was very reliable and cool to see the participants screen update no matter how a user joined, switched roles, or left the stream.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

Simulator Screen Shot - iPhone 13 Pro - 2021-11-09 at 16 12 35


apps/ios/LiveVideo/LiveVideo/Views/Participants/ViewersViewModel.swift
```diff
@@ -0,0 +1,114 @@
...
+            userIdentity = viewer.userIdentity
+        }
+        
+        init(raisedHand: RaisedHandsStore.RaisedHand) {
+            userIdentity = raisedHand.userIdentity
+        }
+    }
+    
+    @Published var viewersWithRaisedHand: [Viewer] = []
+    @Published var viewersWithoutRaisedHand: [Viewer] = []
```
timrozum, line 22: The UX design calls for displaying all the viewers in a list with the raised hands at the top.

Long story short we must use separate sync maps for viewers and raised hands because of the way sync works.

The viewers map contains all viewers including viewers that may have their hand raised. And so the client must filter users out of the viewers map if they have their hand raised. Why not have the server do this filtering?

1. It seemed more general purpose and UI agnostic to simply have a set of all viewers and then the client can do whatever they want with it.
2. We would have to rename the viewers map to something like "viewers without hand raised" which is fine but a bit clunky.
3. If the server did the filtering the client would not be able to animate list updates in the UI because the updates to the viewers map and raised hands map would come in at different times. And even worse we would see strange things like a user showing up twice or not at all momentarily. In my next ticket we will see issues like this when the user switches between viewers and speakers but I think that won't be bad because the user is switching between different list sections in the UI as opposed to changing position inside one section. It seems like the latter would look like a bug if not done well. Or is it just me?

Make sense? Thoughts?

apps/ios/LiveVideo/LiveVideo/Twilio/Sync/ViewersStore.swift
```diff
@@ -0,0 +1,86 @@
+//
+//  Copyright (C) 2021 Twilio, Inc.
+//
+
+import TwilioSyncClient
+import Combine
+
+class ViewersStore: NSObject, SyncStoring, ObservableObject {
```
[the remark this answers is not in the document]
timrozum: I ended up pushing changes to improve code reuse.

apps/ios/LiveVideo/LiveVideo/Views/Participants/ViewersViewModel.swift
```diff
@@ -0,0 +1,98 @@
+//
+//  Copyright (C) 2021 Twilio, Inc.
+//
+
+import Combine
+
+class ViewersViewModel: ObservableObject {
+    @Published var viewersWithRaisedHand: [SyncUsersStore.User] = []
+    @Published var viewersWithoutRaisedHand: [SyncUsersStore.User] = []
```
timrozum, line 9: The UX design calls for displaying all the viewers in a list with the raised hands at the top.

Long story short we must use separate sync maps for viewers and raised hands because of the way sync works.

The viewers map contains all viewers including viewers that may have their hand raised. And so the client must filter users out of the viewers map if they have their hand raised. Why not have the server do this filtering?

1. It seemed more general purpose and UI agnostic to simply have a set of all viewers and then the client can do whatever they want with it.
1. If the server did the filtering the client would not be able to animate list updates in the UI because the updates to the viewers map and raised hands map would come in at different times. And even worse we would see strange things like a user showing up twice or not at all momentarily. In my next ticket we will see issues like this when the user switches between viewers and speakers but I think that won't be bad because the user is switching between different list sections in the UI as opposed to changing position inside one section. It seems like the latter would look more like a bug if not done well compared to the former. Or is it just me?

Make sense? Thoughts?

apps/ios/LiveVideo/LiveVideo/Twilio/Sync/SyncUsersStore.swift
```diff
@@ -5,25 +5,24 @@
...
-    private var newRaisedHands: [RaisedHand] = [] {
-        didSet {
-            haveNew = !newRaisedHands.isEmpty
-        }
-    }
+    let userAddedPublisher = PassthroughSubject<User, Never>()
+    let userRemovedPublisher = PassthroughSubject<User, Never>()
     var uniqueName: String!
     var errorHandler: ((Error) -> Void)?
+    private(set) var users: [User] = []
```
aaalaniz, line 25: Nit, but perhaps would a Set would work better here because the order of users is not useful in this layer.

apps/ios/LiveVideo/LiveVideo/Views/Participants/ViewersViewModel.swift
timrozum: I'm feeling good about this and will resolve.

aaalaniz: This approach is reasonable to me. A raised hand user __is a__ viewer so this is accurate IMO.

serverless/functions/create-stream.js
```diff
@@ -191,6 +191,41 @@ module.exports.handler = async (context, event, callback) => {
     return callback(null, response);
   }
 
+  const viewersMapName = `viewers`
+  
+  // Create viewers map
+  try {
+    raisedHandsMap = await streamSyncClient.syncMaps.create({
```
aaalaniz, line 198:
```suggestion
    viewersMap = await streamSyncClient.syncMaps.create({
```

serverless/functions/join-stream-as-speaker.js
```diff
@@ -207,7 +223,9 @@ module.exports.handler = async (context, event, callback) => {
   response.setBody({
     token: token.toJwt(),
     sync_object_names: {
+      viewers_map: 'viewers',
       raised_hands_map: `raised_hands`,
+      viewer_document: `viewer-${user_identity}`,
```
aaalaniz, line 228: I know we use this convention right now, but is the `viewer-` prefix required?

serverless/functions/rooms-webhook.protected.js
```diff
@@ -2,13 +2,50 @@
...
+          message: 'error getting stream sync client',
+          explanation: e.message,
+        },
+      });
+      return callback(null, response);
+    }
+
+    // Remove user from viewers map
+    try {
+      await streamSyncClient.syncMaps('viewers').syncMapItems(event.ParticipantIdentity).remove();
```
aaalaniz, line 32: Do we also need to remove them from the raisedHands map?

serverless/functions/viewer-connected-to-player.js
```diff
@@ -0,0 +1,93 @@
...
+  const client = context.getTwilioClient();
+
+  let room, streamSyncClient;
+
+  try {
+    // See if a room already exists
+    room = await client.video.rooms(stream_name).fetch();
+  } catch (e) {
+    console.error(e);
+    response.setStatusCode(500);
```
aaalaniz, line 43: Nit, but thoughts on returning 404 here?

aaalaniz (commented):
Looks great!

apps/ios/LiveVideo/LiveVideo/Twilio/Sync/SyncUsersStore.swift
timrozum: Good question but I think it's probably best to keep an array. SwiftUI needs an array. And sync provides an array. And so it would make for a lot of converting between set and array.

serverless/functions/join-stream-as-speaker.js
timrozum: We could technically get away without the prefix for this file now but it may get us in the future if we need another user specific sync object. And I think it's easier to follow our sync design if all objects have a descriptive name. It will be easier to refer to in documentation I think. So I think it's a good convention.

serverless/functions/rooms-webhook.protected.js
timrozum: For this PR this is handled in the `join-stream-as-speaker` endpoint because that is where it existed previously and it works. But in the next PR that adds the speakers map I did move it to this webhook because it is a better time to do it.

serverless/functions/viewer-connected-to-player.js
timrozum: That is logical to me but currently the backend only uses 400 and 500 error codes. Maybe it is useful to reserve 404 specifically for invalid URL. I know I run into 404 sometimes when testing and then immediately realize oh I need to update the backend URL. I could also see this being a 400 instead of 500. It probably isn't a backend error. I can go either way just some things to think about.

serverless/functions/create-stream.js
timrozum: I pushed changes for this.

aaalaniz (approved):

apps/ios/LiveVideo/LiveVideo/Launch/LiveVideoApp.swift
```diff
@@ -25,14 +25,20 @@ struct LiveVideoApp: App {
...
+                .environmentObject(viewersViewModel)
                 .environmentObject(api)
                 .onAppear {
                     let localParticipant = LocalParticipantManager(authManager: authManager)
                     let roomManager = RoomManager()
                     roomManager.configure(localParticipant: localParticipant)
                     let viewerStore = ViewerStore()
-                    let syncManager = SyncManager(raisedHandsStore: raisedHandsStore, viewerStore: viewerStore)
+                    let raisedHandsStore = SyncUsersStore()
+                    let viewersStore = SyncUsersStore()
```
piyushtank, line 36: Assuming raisedHandsStore and viewersStore doesn't share any participants

piyushtank (approved):

apps/ios/LiveVideo/LiveVideo/Launch/LiveVideoApp.swift
timrozum: They do share some participants. `raisedHandsStore` is all users that have their hand raised. `viewersStore` is all viewers regardless if hand is raised or not. 

Some more background on the sync architecture. We really need to use a separate sync map for raised hands. There could be 1k users in the viewers sync map. We can't just add a hand raised attribute to users in the `viewers` map because of pagination. We would have to paginate through all users just to find the users that have their hand raised to display in the UI. And all that pagination doesn't scale and isn't feasible. So because of how sync works we must use a separate map just for raised hands. If sync had a way to subscribe to queries that would have helped but it does not.

Also some kind of related discussion here: https://github.com/twilio/twilio-live-interactive-video/pull/55#discussion_r746805646

timrozum: The link I added above tries to explain why we wouldn't have the server manage these two maps such that there is no overlap in users.

PR #56, opened 2021-11-10
VIDEO-7601: Add speakers to participants screen

timrozum:
### JIRA link(s):

- [VIDEO-7601](https://issues.corp.twilio.com/browse/VIDEO-7601)

### Description

1. Make backend changes to create a speakers map and add/remove speakers when they connect/disconnect to the video room.
2. Add speakers to the participants screen in the iOS app. It is not exactly to the UX spec but it is close and this PR is more focused on making sync work.
3. In the iOS app I moved the code from `ViewersViewModel` into `ParticipantsViewModel`. Originally I kept them separate because I thought it would be too much to combine but I think it's fine and most straightforward for customers.
4. Only show the viewer connected welcome alert once per stream because it doesn't really make sense to show when a user moves from speakers back to viewers.
5. You may see in the video that the player view flashes sometimes. I don't think that is related to this PR. I think it may be an app issue and I need to look into it.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

https://user-images.githubusercontent.com/1930363/141205440-bdcf11da-7031-4c66-9ab8-bfcb860e1974.mov


serverless/functions/sync-webhook.protected.js
```diff
@@ -24,6 +24,23 @@ exports.handler = async function (context, event, callback) {
       return callback(null, response);
     }
 
+    // Remove user from speakers map
+    try {
+      await streamSyncClient.syncMaps('speakers').syncMapItems(event.Identity).remove();
```
aaalaniz, line 29: It it every possible for this event to occur but the user is still connected to the Room?

serverless/functions/join-stream-as-speaker.js
```diff
@@ -129,24 +145,6 @@ module.exports.handler = async (context, event, callback) => {
     return callback(null, response);
   }
 
-  // Lower the participant's hand
-  try {
```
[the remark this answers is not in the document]
timrozum: I think it is simplest to modify these user sync maps from the webhooks for consistency. It's easier to keep things straight. And it's arguably better UX. Updates are made to the viewers and speakers maps at the same time. And so we minimize any gap that the UI would have moving a viewer to speaker. 

I can see an argument that it doesn't completely match reality but that seems minor. It may make sense to reconsider when we the player streamer callback adds a disconnect event.

serverless/functions/sync-webhook.protected.js
timrozum: Sync has a 5 second timeout and rooms has a 30 second timeout I think. So if something goes wrong yeah the user could probably linger in the room a little longer than in sync. 

We could use the rooms REST API to remove the user immediately. I could go either way on this. It would add a little complexity and it's an edge case but it would make rooms and sync more consistent for the client. If closing web tabs happens a lot it may be worth it.

serverless/functions/create-stream.js
```diff
@@ -157,6 +157,41 @@ module.exports.handler = async (context, event, callback) => {
     return callback(null, response);
   }
 
+  const speakersMapName = `speakers`
+  
+  // Create speakers map
+  try {
+    raisedHandsMap = await streamSyncClient.syncMaps.create({
```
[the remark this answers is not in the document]
timrozum: Pushed changes for this.

serverless/functions/sync-webhook.protected.js
aaalaniz: I think it depends on if we think losing sync connectivity is a terminal state. I think this approach is fine

aaalaniz (approved):

piyushtank (approved):

PR #59, opened 2021-11-16
VIDEO-7753: Add host state to sync

timrozum:
### JIRA link(s):

- [VIDEO-7753](https://issues.corp.twilio.com/browse/VIDEO-7753)

### Description

1. Make backend changes to store host status in speakers map.
2. Add host label to speakers grid and participants screen in the iOS app.
4. Add participants button to viewer toolbar in iOS app. This should have been done when we started showing viewers and speakers on the participants screen.
5. The video composer does not label hosts yet. We should try to make sure that the changes in this PR will play nice with how we think this will be done in the future.
6. No breaking changes with the web app.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

Add host label to speakers grid:
Simulator Screen Shot - iPhone 13 - 2021-11-16 at 11 30 14

Add host label to participants screen:
Simulator Screen Shot - iPhone 13 - 2021-11-16 at 11 29 54

Add participants button to viewer toolbar:
Simulator Screen Shot - iPhone 13 Pro - 2021-11-16 at 11 27 22


apps/ios/LiveVideo/LiveVideo/Views/Stream/SpeakerGridViewModel.swift
```diff
@@ -102,4 +110,20 @@ class SpeakerGridViewModel: ObservableObject {
...
+    private func makeSpeaker(participant: LocalParticipantManager) -> SpeakerVideoViewModel {
+        let isHost = speakersStore.host?.identity == participant.identity
+        return SpeakerVideoViewModel(participant: participant, isHost: isHost)
+    }
+
+    private func makeSpeaker(participant: RemoteParticipantManager) -> SpeakerVideoViewModel {
+        let isHost = speakersStore.host?.identity == participant.identity
+        return SpeakerVideoViewModel(participant: participant, isHost: isHost)
+    }
+}
```
timrozum, line 123: I am for sure stretching the existing architecture here to combine room and sync state. I was able to avoid adding another user abstraction for now. I suspect when I add the screen share layout a new user abstraction to help merge state from two sources may be unavoidable.

I already have `RemoteParticipantManager` and `LocalParticipantManager` for room state. `SyncUsersStore.User` for sync state. And `SpeakerVideoViewModel` for UI state. To really do this right I would probably need to create a new user abstraction with merged state that sits between `SpeakerVideoViewModel` and the other room and sync abstractions. But it is a new layer just for the single host property. So I held off for now by bending rules in a couple places.

apps/ios/LiveVideo/LiveVideo/Views/Participants/ParticipantsViewModel.swift
```diff
@@ -149,3 +153,9 @@ class ParticipantsViewModel: ObservableObject {
         viewerCount = viewersWithoutRaisedHand.count + viewersWithRaisedHand.count
     }
 }
+
+extension SyncUsersStore.User {
+    var displayName: String {
```
aaalaniz, line 158: Can this be `let` instead of `var` ?

aaalaniz (approved):

apps/ios/LiveVideo/LiveVideo/Views/Participants/ParticipantsViewModel.swift
timrozum: It cannot because it is a computed property. Only stored properties can be let. However because it is a stored property that is only read only it has the same exact behavior as a let in this case.

PR #61, opened 2021-11-17
Add passcode

olipyskoty:
**Contributing to Twilio**

### JIRA link(s):

- [VIDEO-7083](https://issues.corp.twilio.com/browse/VIDEO-7083)

### Description

This PR provides passcode functionality using Twilio Serverless functions. This implementation ensures that passcodes are required, and participants will have the option to "log out" if they please. The passcode is sent in each request's headers. The frontend UI is copied over from the Twilio Video React app. This PR also removes a lot of the firebase and Twilio Programmable Video `token` logic. Please let me know if I accidentally removed something that is required.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


timmydoza:
Still need to review everything, but let's update `serverless/scripts/list.js` so that it includes the passcode in the URL. So the output will look like this:

```
App deployed to: https://twilio-live-interactive-video-3456-4567-dev.twil.io?passcode=12323434564567
Passcode: 123 234 3456 4567
This URL is for demo purposes only. It will expire on Wed Nov 24 2021 16:56:13 GMT-0700 (Mountain Standard Time)
```

Can probably find the code for this in the react app. I think the app itself already consumes this parameter, so we just need to update what's printed in the console. 

Also, does `serverless/functions/viewer-connected-to-player.js` need the auth handler called in it?

apps/web/src/components/PrivateRoute/PrivateRoute.tsx
```diff
@@ -6,7 +6,7 @@ import { useAppState } from '../../state';
 export default function PrivateRoute({ children, ...rest }: RouteProps) {
   const { isAuthReady, user } = useAppState();
 
-  const renderChildren = user || !process.env.REACT_APP_SET_AUTH;
+  const renderChildren = user;
```
timmydoza, line 9: This looks kinda funny. Wonder if we could just do:

`if (!user && !isAuthReady)`

apps/web/src/state/index.tsx
```diff
@@ -1,156 +1,51 @@
...
-        return res;
-      })
-      .catch(err => {
-        setError(err);
-        setIsFetching(false);
-        return Promise.reject(err);
-      });
+  contextValue = {
+    ...contextValue,
+    ...usePasscodeAuth(),
```
timmydoza, line 45: Here I think we can move `...usePasscodeAuth()` up to the first `contextValue` declaration. 

```
  const contextValue = {
    error,
    setError,
    activeSinkId,
    setActiveSinkId,
    settings,
    dispatchSetting,
    appState,
    appDispatch,
    ...usePasscodeAuth(),
  } as StateContextType;

  return <StateContext.Provider value={contextValue}>{props.children}</StateContext.Provider>;
  ```
  
  Really nice cleanup in this file!!

File: apps/web/src/state/usePasscodeAuth/usePasscodeAuth.ts [hunk 3]
```diff
@@ -1,50 +1,32 @@
 import { useCallback, useEffect, useState } from 'react';
 import { useHistory, useLocation } from 'react-router-dom';
 import { Location } from 'history';
+import { apiClient } from '../api/api';
 
-const endpoint = process.env.REACT_APP_TOKEN_ENDPOINT || '/token';
+const endpoint = '/verify-passcode';
```
timmydoza, line 6: Don't think we need a variable for this. I think this value is just used in one spot. 

`return fetch('verify-passcode', {`

serverless/functions/verify-passcode.js
```diff
@@ -0,0 +1,18 @@
...
+module.exports.handler = async (context, event, callback) => {
+  const authHandler = require(Runtime.getAssets()['/auth.js'].path);
+  authHandler(context, event, callback);
+
+  let response = new Twilio.Response();
+  response.appendHeader('Content-Type', 'application/json');
+
+  response.setStatusCode(200);
+
+  response.setBody({});
```
timmydoza, line 15: Not sure this will actually mean anything, but I'm wondering if we could do something like this:
`response.setBody({ verified: true });`

Some of the other endpoints here have something similar.

timmydoza (changes requested):
Nice work! Just have a few tiny comments. 

Thanks for the cleanup in this PR! There's definitely still more to be done. Plus we can remove whole files that are no longer used (like `useFirebase.ts`). But I think we can do all this in a new PR since we already have a ticket for cleaning up everything.

serverless/middleware/auth.private.js
```diff
@@ -0,0 +1,38 @@
+/* global Twilio */
+'use strict';
+
+module.exports = async (context, event, callback) => {
+  const { PASSCODE, APP_EXPIRY, DOMAIN_NAME } = context;
+
+  const passcode = event.request.headers.authorization;
```
timrozum, line 7: @olipyskoty @timmydoza what do you think about adding user identity to the header also? So we don't have to separately add user identity as a parameter to requests? 

One way to do that would be using "basic" authorization which contains username and password. Here is an example of how this is encoded in swift:
```swift
    /// Returns a `Basic` `Authorization` header using the `username` and `password` provided.
    ///
    /// - Parameters:
    ///   - username: The username of the header.
    ///   - password: The password of the header.
    ///
    /// - Returns:    The header.
    public static func authorization(username: String, password: String) -> HTTPHeader {
        let credential = Data("\(username):\(password)".utf8).base64EncodedString()

        return authorization("Basic \(credential)")
    }
```
This would be easy on iOS since this networking library would do the encoding for me. I assume there is a JS package to encode and decode.

I guess only 3 out of the 6 endpoints use identity so this isn't really a must. But thought it was worth tossing around.

timmydoza (approved):

serverless/middleware/auth.private.js
timmydoza: I think we could, but it doesn't feel completely right. It seems odd to add the `user_identity` to the Authorization header when it's never going to be used for authentication or authorization. Since the API only uses the `user_identity` parameter to access specific resources in the Twilio API, I think it makes sense that `user_identity` should be in the request body and not the Authorization header. 

Also, I fear that adding Basic Auth might give the impression that this app is more secure than it really is. I wouldn't want to give the impression that one user can't change another user's hand raised status (for example). The passcode is only intended to protect the endpoints from unauthorized use, not to protect each individual user's resources from other malicious users. 

Down the road, I think we could implement more robust authentication that protects users from other users using Bearer Authentication. The Twilio access token that is given to every user can be used for this. The backend is able to tell that it is signed correctly (the BE has access to the right secret for this), the user's identity is embedded in the token, and each user only has access to their token.

timrozum: Very good points thanks!

PR #62, opened 2021-11-17
VIDEO-7752: Rename viewer document

timrozum:
### JIRA link(s):

- [VIDEO-7752](https://issues.corp.twilio.com/browse/VIDEO-7752)

### Description

These are the reasons for the rename:

1. Prevent confusion between similarly named sync objects. We have a `viewers` map and a `viewer-bob` document. It will be easier to describe these objects to customers if they have more distinct names. And so I have renamed to `user-bob` since it is unnecessary to scope this object to viewers only. 
2. Scoping to any user makes the object more general purpose which may be useful in the future. There is an argument that scoping to viewers is more self documenting for how the object is currently used but that seems like a minor tradeoff especially since the contents of the object should be intuitive.
3. In order to simplify sync setup on the clients we were already creating this object for both viewers and speakers when they join a stream. And so it's a little less strange to create a general purpose user object when a speaker joins.

Made changes to backend, iOS, and web. The backend and web changes were small. The iOS changes were larger because I needed to prevent other name confusion internal to iOS.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/apps/ios/LiveVideo/LiveVideo/Launch/LiveVideoApp.swift
+++ b/apps/ios/LiveVideo/LiveVideo/Launch/LiveVideoApp.swift
@@ -29,15 +29,15 @@ struct LiveVideoApp: App {
                     let localParticipant = LocalParticipantManager(authManager: authManager)
                     let roomManager = RoomManager()
                     roomManager.configure(localParticipant: localParticipant)
-                    let viewerStore = ViewerStore()
-                    let speakersStore = SyncUsersStore()
-                    let raisedHandsStore = SyncUsersStore()
-                    let viewersStore = SyncUsersStore()
+                    let userDocument = SyncUserDocument()
+                    let speakersMap = SyncUsersMap()
+                    let raisedHandsMap = SyncUsersMap()
+                    let viewersMap = SyncUsersMap()
                     let syncManager = SyncManager(
-                        speakersStore: speakersStore,
-                        viewersStore: viewersStore,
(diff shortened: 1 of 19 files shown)


apps/ios/LiveVideo/LiveVideo/Launch/LiveVideoApp.swift
```diff
@@ -29,15 +29,15 @@ struct LiveVideoApp: App {
...
                     let roomManager = RoomManager()
                     roomManager.configure(localParticipant: localParticipant)
-                    let viewerStore = ViewerStore()
-                    let speakersStore = SyncUsersStore()
-                    let raisedHandsStore = SyncUsersStore()
-                    let viewersStore = SyncUsersStore()
+                    let userDocument = SyncUserDocument()
+                    let speakersMap = SyncUsersMap()
+                    let raisedHandsMap = SyncUsersMap()
+                    let viewersMap = SyncUsersMap()
```
timrozum, line 35: I had to rename some other things in iOS because I was dealing with `SyncUsersStore` history which is a reusable way to load sync maps that contain users.

During the rename I made things a little less abstract which I maybe wouldn't do in a real production app but I think it will be easier for customers to follow in this app.

timmydoza (approved):

aaalaniz (approved):

PR #63, opened 2021-11-19
VIDEO-7604: Allow host to move speaker to viewers

timrozum:
### JIRA link(s):

- [VIDEO-7604](https://issues.corp.twilio.com/browse/VIDEO-7604)

### Description

1. Add feature for host to move speaker to viewers. The mechanics are the same as the live audio app.
2. Note that the ordering of the items in the menu depends if the system decides to display the menu above or below the view that is tapped. In the code the remove speaker item is last in the menu. But in the video it is displayed first so that it is furthest from the tap location and the menu is displayed above the tap.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

https://user-images.githubusercontent.com/1930363/142695545-6ac965f7-ef49-421b-a6d8-25b22e32dae3.mov

### DIFF
--- a/apps/ios/LiveVideo/LiveVideo/Twilio/Room/RoomManager.swift
+++ b/apps/ios/LiveVideo/LiveVideo/Twilio/Room/RoomManager.swift
@@ -21,6 +21,7 @@ class RoomManager: NSObject {
     // MARK: -
 
     var roomSID: String? { room?.sid }
+    var roomName: String? { room?.name }
     private(set) var localParticipant: LocalParticipantManager!
     private(set) var remoteParticipants: [RemoteParticipantManager] = []
     private var room: Room?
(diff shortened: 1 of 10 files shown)


apps/ios/LiveVideo/LiveVideo/Twilio/Room/RoomManager.swift
```diff
@@ -80,14 +81,18 @@ extension RoomManager: RoomDelegate {
...
+            } else if (error as NSError).isParticipantNotFoundError {
+                // Can receive this error when a speaker is removed by host if there is other activity in progress
+                handleError(LiveVideoError.speakerMovedToViewersByHost)
+            } else {
+                handleError(error)
+            }
         } else {
-            handleError(error)
+            // Most of the time there is no error when speaker is removed by host
+            handleError(LiveVideoError.speakerMovedToViewersByHost)
```
piyushtank, line 95: Since this is not an error condition, should it we use different route instead?

apps/ios/LiveVideo/LiveVideo/Twilio/Stream/StreamViewModel.swift
```diff
@@ -106,11 +107,32 @@ class StreamViewModel: ObservableObject {
...
+    }
+
+    var isStreamEndedByHostError: Bool {
+        if case .streamEndedByHost = self as? LiveVideoError {
+            return true
+        }
+
+        return false
+    }
+}
```
piyushtank, line 138: These are not errors, they are reflection of  admin actions. IMHO

timrozum: Yeah I get how one can interpret it that way.

I think it is a gray area. These are fatal errors from `RoomManager` in the sense that the user has been unexpectedly disconnected from the room. And also these cases are derived by errors received from the video SDK. I'm just transforming them a little for this app.

I guess the best alternative would be to create 2 new publishers instead of reusing `roomDisconnectPublisher`. But now there are multiple "disconnect" publishers which seems more complex and problematic. Or instead of `roomDisconnectPublisher` sending an `Error` I could create a new type that can hold an error or other information. But that seems unconventional and will also complicate usage.

If these states were received from `Sync` instead of the video SDK and "events" were sent instead of errors I think we would not use errors for this. But for a number of reasons we are receiving this state from the video SDK as errors.

So I totally get your perspective and it's good feedback. But I think all things considered using errors is the best solution because it is much simpler, the state is derived from video SDK errors, and I wouldn't go so far as to say using errors for these events is incorrect or confusing.

piyushtank (approved):

PR #64, opened 2021-11-23
Fix media issue on event join

timmydoza:
**Contributing to Twilio**

### JIRA link(s):

- N/A

### Description

This pull request prevents users from joining an event before their media devices have been acquired. Previously, users could join a room while the device permissions window was still pending, which meant that they were not able to publish audio or video in the event. Now, the "Join Event" button is disabled until the users audio and video tracks have been acquired.

image

### Before review
* [ ] Verified locally with all effected platforms
* [ ] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


olipyskoty (approved):
this works 🎉 suppose we may need to update some tests when we get around to it.

PR #65, opened 2021-12-01
VIDEO-7121: Add passcode support to iOS app

timrozum:
### JIRA link(s):

- [VIDEO-7121](https://issues.corp.twilio.com/browse/VIDEO-7121)

### Description

Added passcode support to the iOS app so that we can control access better and change the backend URL without having to build and distribute the iOS app again. The backend URL is derived from the passcode.

1. Add new screen for user to enter passcode when they sign in.
2. Use the passcode components to make the backend URL.
3. Add the passcode to the HTTP header for all reference backend requests.
3. Use `verify-passcode` endpoint to verify that the passcode is correct.
4. Store the passcode in the keychain if it is correct.
5. Enable code coverage.

Will handle readme changes separately since we need to address that as a team.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

Simulator Screen Shot - iPhone 13 - 2021-12-01 at 14 46 35
Simulator Screen Shot - iPhone 13 - 2021-12-01 at 14 46 50


apps/ios/LiveVideo/LiveVideo/Views/SignIn/EnterUserIdentityView.swift
```diff
@@ -0,0 +1,60 @@
...
+                .background(
+                    /// Set the `NavigationLink` as background with 0 opacity so it is completely invisible and doesn't alter spacing
+                    NavigationLink(
+                        destination: EnterPasscodeView(userIdentity: userIdentity),
+                        isActive: $isShowingPasscodeView
+                    ) {
+                        EmptyView()
+                    }
+                        .opacity(0)
+                )
```
timrozum, line 49: These are the main changes in this file. It is not a new file but git thinks it is.

apps/ios/LiveVideo/LiveVideo/Managers/Auth/AuthManager.swift
```diff
@@ -3,21 +3,61 @@
...
 
-    func signIn(userIdentity: String) {
-        Defaults.userIdentity = userIdentity
-        self.userIdentity = userIdentity
-        isSignedOut = false
+    func configure(api: API) {
+        self.api = api
+
+        if let userIdentity = keychain[userIdentityKey], let passcode = keychain[passcodeKey] {
+            isSignedOut = false
```
piyushtank, line 20: Assuming this is always used from main thread

apps/ios/LiveVideo/LiveVideo/Managers/Auth/AuthManager.swift
```diff
@@ -3,21 +3,61 @@
...
+            guard let self = self else {
+                return
+            }
+            
+            switch result {
+            case .success:
+                self.keychain[self.userIdentityKey] = userIdentity
+                self.keychain[self.passcodeKey] = passcode
+                self.userIdentity = userIdentity
+                self.isSignedOut = false
```
piyushtank, line 44: Is this called on main thread?

piyushtank: yes, tested. Its called on the main thread so we are good here. worth adding a comment I guess.

piyushtank (commented):
Getting following errors when I tried to create event -

```
2021-12-01 17:16:45.880377-0800 LiveVideo[3678:606224] Can't find keyplane that supports type 4 for keyboard iPhone-PortraitChoco-NumberPad; using 27059_PortraitChoco_iPhone-Simple-Pad_Default
2021-12-01 17:17:07.056756-0800 LiveVideo[3678:606224] [Presentation] Attempt to present <_TtGC7SwiftUI29PresentationHostingControllerVS_7AnyView_: 0x15be45f60> on <_TtGC7SwiftUI19UIHostingControllerGVS_15ModifiedContentVS_7AnyViewVS_12RootModifier__: 0x15bf0cd00> (from <_TtGC7SwiftUI19UIHostingControllerGVS_15ModifiedContentGS1_VVS_22_VariadicView_Children7ElementVS_24NavigationColumnModifier_GVS_18StyleContextWriterVS_19SidebarStyleContext___: 0x15bf1b970>) which is already presenting <_TtGC7SwiftUI29PresentationHostingControllerVS_7AnyView_: 0x15d131000>.
2021-12-01 17:17:21.801545-0800 LiveVideo[3678:606224] [SwiftUI] Currently, only presenting a single sheet is supported.
The next sheet will be presented when the currently presented sheet gets dismissed.
2021-12-01 17:17:23.829065-0800 LiveVideo[3678:606224] [SwiftUI] Currently, only presenting a single sheet is supported.
The next sheet will be presented when the currently presented sheet gets dismissed.
2021-12-01 17:17:55.829038-0800 LiveVideo[3678:606553] [tcp] tcp_input [C1.1:3] flags=[R] seq=2866353748, ack=0, win=0 state=LAST_ACK rcv_nxt=2866353748, snd_una=4048205600
2021-12-01 17:17:55.829502-0800 LiveVideo[3678:606553] [tcp] tcp_input [C1.1:3] flags=[R] seq=2866353748, ack=0, win=0 state=CLOSED rcv_nxt=2866353748, snd_una=4048205600
2021-12-01 17:17:55.829673-0800 LiveVideo[3678:606553] [tcp] tcp_input [C1.1:3] flags=[R] seq=2866353748, ack=0, win=0 state=CLOSED rcv_nxt=2866353748, snd_una=4048205600
```

apps/ios/LiveVideo/LiveVideo/Managers/Auth/AuthManager.swift
timrozum: I think in the app layer it is probably common to assume completion closures are called on the main queue unless otherwise stated in order to keep concurrency as simple as possible. But to help with this I added a new comment inside `API.swift` that explains that all API responses are called on the main queue. I think it's probably best to explain this at the API level instead of the various places API requests are made.

timrozum:
@piyushtank I pushed a fix for that intermittent UI navigation bug. It's kind of a hack but it is reliable and simple and I think the root cause is probably a SwiftUI bug. 

In the future I may try to refactor how some of the navigation is done on this screen and see if that helps. But I think that is a separate effort since this screen is central to the navigation in this app and could take some time.

piyushtank (commented):
When I delete the app, is it expected to remember the credentials ? Just want to make sure that its expected.

timrozum:
@piyushtank yes it is expected that the credentials are still there even after an app delete. This is just how the keychain works which I agree is kind of unexpected. But it is how all the apps I have worked on behave. 

I can imagine a way to add logic that clears out the keychain at first app launch but that is more state to keep track off and could do more harm than good. And most of the time it is a good thing for users that they don't have to sign in again.

piyushtank (approved):

PR #77, opened 2022-01-12
Video 8216 add mobile styling

olipyskoty:
**Contributing to Twilio**

### JIRA link(s):

- [VIDEO-8216](https://issues.corp.twilio.com/browse/VIDEO-8216)

### Description

This PR improves the mobile styling for the live interactive web app. All input fields now automatically focus, as well as some of the buttons. Note, the autofocus for the buttons didn't seem to always work on Chrome for me? I also think we will need to test this on a couple of different devices to confirm the styling looks okay.

https://user-images.githubusercontent.com/77076398/149047352-bddb6467-8785-4a22-bbe6-c97c05f3edad.mp4

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


olipyskoty:
@timmydoza I updated the buttons to use `<BaseButton />` and brought back the `<Paper />` elements. One thing I noticed is that although they are "tab-able" now, pressing Enter or Space doesn't seem to activate the button and I'm not sure why. I thought maybe it was because of [this](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#:~:text=Note%3A%20If%20using%20role%3D%22button%22%20instead%20of%20the%20semantic%20%3Cbutton%3E%20or%20%3Cinput%20type%3D%22button%22%3E), but Chrome dev tools shows that they are `<buttons>`?

Also, I found that we probably need to update the `<MenuBar />` and `<PlayerMenuBar />` mobile styling for the Viewer and Speaker (at least for Safari):
IMG_7617
IMG_7618

timmydoza:
> pressing Enter or Space doesn't seem to activate the button

I think this has to do with the fact that the `onClick` listener is on the Paper component and not the ButtonBase component. So in the dom, the listener is on the div inside the button, and not the button itself. It works with this:

```
<ButtonBase focusRipple onClick={() => dispatch({ type: 'set-participant-type', participantType: 'host' })}>
```

timmydoza (changes requested):
Looks good! Almost there. Let's:
- Move the onClick listeners from the Paper component to the ButtonBase component.
- Remove the `focusVisible` and `disabledPaper` classes - doesn't look like they are being used anywhere.
- Make a ticket to fix the mobile menu bar in this and the React app. Looks like the React app has the same issue. I think we should fix it, but as a separate PR.

timmydoza (approved):

PR #87, opened 2022-01-26
Improve stream end so the backend does not log useless errors

timrozum:
Rely on the rooms webhook to clean up resources when the room ends. And then anywhere we want to end the stream just end the room. The rooms webhook will handle the rest.

There was some duplicate clean up that was happening which caused the backend to log errors which generated noise.

Tested the changes and resources are cleaned up and I did not see any errors in the Twilio console.


timmydoza (approved):
Works well!

PR #94, opened 2022-02-11
iOS add presentation layout

timrozum:
When a speaker share's their screen change the video layout to show the screen share video and the last dominant speaker.

The change between grid layout and presentation layout is automatic. The user cannot change layout manually.

This only works when a user shares their screen from the web app. This iOS app does not yet allow users to share their screen from iOS.

When in presentation layout the host cannot tap the speaker to access the moderation menu. I will fix this in a separate ticket so the menu works on both the grid layout and presentation layout. We wanted to make a small UX change to this anyway to make the feature more discoverable and if I make the UX change it will also make it easier to add this feature to the presentation layout.

Simulator Screen Shot - iPhone 13 Pro - 2022-02-11 at 15 43 15
Simulator Screen Shot - iPhone 13 Pro - 2022-02-11 at 15 43 46


apps/ios/LiveVideo/LiveVideo/Views/Stream/PresentationLayout/PresentationLayoutViewModel.swift
```diff
@@ -0,0 +1,99 @@
...
+class PresentationLayoutViewModel: ObservableObject {
+    struct Presenter {
+        var identity: String = ""
+        var displayName: String = ""
+        var presentationTrack: VideoTrack?
+    }
+
+    @Published var isPresenting = false
+    @Published var presenter = Presenter()
+    @Published var dominantSpeaker = SpeakerVideoViewModel()
```
timrozum, line 17: It is a little odd to default these to basically empty values but I think it is the best option. SwiftUI binding does not play nice with optional chaining. And so if these properties are optional it creates all sorts of issues in the UI when I try to bind some property of a struct to some view, because I have to unwrap the struct to get to the property. So instead these properties are not optional but the structs may basically have empty data inside them. The UI should rely on `isPresenting` to determine if these other properties have data that should be displayed.

We can talk offline about other solutions I thought of as well.

apps/ios/LiveVideo/LiveVideo/Views/Stream/SpeakerVideoViewModel.swift
```diff
@@ -6,22 +6,20 @@ import TwilioVideo
...
-    let dominantSpeakerStartTime: Date
+    var identity = ""
+    var displayName = ""
+    var isYou = false
+    var isMuted = false
+    var dominantSpeakerStartTime: Date = .distantPast
+    var isDominantSpeaker = false
     var cameraTrack: VideoTrack?
-    var shouldMirrorCameraVideo: Bool
+    var shouldMirrorCameraVideo = false
```
timrozum, line 16: With SwiftUI I'm finding it often makes sense to use vars because:

1. It must be a var to bind the value to a view.
2. In a situation like `PresentationLayoutViewModel` where I need "empty" values it is convenient to use default values as done here.

At times SwiftUI can feel a little less type safe and also like there is a little duplicate state in the model to expose everything properly to the UI. But a small price to pay for the benefits of SwiftUI.

apps/ios/LiveVideo/LiveVideo/Views/Stream/SpeakerVideoViewModelFactory.swift
```diff
@@ -0,0 +1,23 @@
+//
+//  Copyright (C) 2022 Twilio, Inc.
+//
+
+import Foundation
+
+class SpeakerVideoViewModelFactory {
```
timrozum, line 7: This code used to be in `SpeakerGridViewModel` but I moved it out so that it could also be used by `PresentationLayoutViewModel`.

If this was a production app it would probably make sense to create an entirely new layer that is above `RoomManager` and below the view models. Such layer would be responsible for merging state from the video SDK and the Sync SDK into a single view of the speakers. But if this was a production app there would probably be a lot more state than just `host` to merge with the video room data. I decided against adding the extra layer because it would add quite a bit of complexity and may be harder for customers to follow. The new layer would probably need to maintain state for all the speakers and have publishers to send out updates.

apps/ios/LiveVideo/LiveVideo/Views/Stream/PresentationLayout/PresentationLayoutViewModel.swift
wolfspy: This is how I set up my UIKit view models usually anyways to prevent them from being optional.

timrozum: Good to know thanks!

PR #108, opened 2022-03-29
iOS app recording feature

timrozum:
1. Allow the host to enable recording when creating a stream.
2. Display recording badge for speakers and viewers when the stream is being recorded.
3. Display recording alert when speakers join a stream that is being recorded.
4. If there is a recording error, display it to the host but don't end the stream.
5. Instead of getting sync object names from the reference backend, just hard-code them in the app. This will make the reference app simpler. We were fetching the sync object names from the reference backend because we expected the names to be more complex. But we changed our sync architecture and it made the object names simpler. 
6. Update `TwilioLivePlayer` package.

### Demo:
https://user-images.githubusercontent.com/1930363/160698057-d1f3fef1-06af-4c93-b9d9-6c3a96ee92d8.mov


apps/ios/LiveVideo/LiveVideo/Twilio/Sync/SyncStreamDocument.swift
```diff
@@ -0,0 +1,113 @@
...
+            
+            self.recording = Recording(isRecording: isRecording, error: error)
+        }
+    }
+        
+    @Published var isRecording = false
+    
+    /// Record errors are informative and do not end the stream, so provide a special path to handle them. It would be nice to have
+    /// a single path for all errors but that would require a significant refactor and probably make other code more complex.
+    @Published var recordError: Error?
```
timrozum, line 35: Just keeping this simple by creating a separate path for record errors because:

1. This is the first time we are pushing errors from the reference backend to clients via sync
2. The recording errors are just informative and not fatal which is different from other backend and SDK errors.
3. There is also some extra logic around recording errors so that we don't show them until after the user is connected to the stream. And so handling state for this error is more complex than other errors that are always immediately handled.

We could evolve this in the future if we have more errors like this.

apps/ios/LiveVideo/LiveVideo/Views/Stream/RecordingBadge.swift
```diff
@@ -0,0 +1,40 @@
+//
+//  Copyright (C) 2022 Twilio, Inc.
+//
+
+import SwiftUI
+
+struct RecordingBadge: View {
```
timrozum, line 7: Stole this from the video app.

apps/ios/LiveVideo/LiveVideo/Views/StreamConfigFlow/EnterStreamNameView.swift
```diff
@@ -8,59 +8,75 @@ struct EnterStreamNameView: View {
...
-
-                TextField("Event name", text: $streamName)
-                    .textFieldStyle(FormTextFieldStyle())
-                    .autocapitalization(.none)
-                    .disableAutocorrection(true)
-
-                VStack {
-                    NavigationLink(destination: SelectRoleView(), isActive: $isShowingSelectRole) {
-                        EmptyView()
+            Form {
```
timrozum, line 42: Realized that SwiftUI has a form built in so going to try to use it for input screens like this. It looks quite nice.

apps/ios/LiveVideo/LiveVideo/Views/StreamConfigFlow/EnterStreamNameView.swift
```diff
@@ -8,59 +8,75 @@ struct EnterStreamNameView: View {
...
+                            }
+                            .buttonStyle(PrimaryButtonStyle(isEnabled: !streamName.isEmpty))
+                            .disabled(streamName.isEmpty)
+                            .padding(.horizontal, -20)
                         }
-                    }
-                    .buttonStyle(PrimaryButtonStyle(isEnabled: !streamName.isEmpty))
-                    .disabled(streamName.isEmpty)
+                ) {
+                    /// Using the footer to have more control over style
```
piyushtank, line 79: Is this intended ?

apps/ios/LiveVideo/LiveVideo/Twilio/Sync/SyncStreamDocument.swift
```diff
@@ -0,0 +1,113 @@
...
+    
+    private func handleError(_ error: Error) {
+        disconnect()
+        errorHandler?(error)
+    }
+    
+    private func update(data: [String: Any], previousData: [String: Any]? = nil) throws {
+        let data = try DocumentData(data: data)
+
+        isRecording = data.recording.isRecording
```
piyushtank, line 75: Is this thread safe here ? same for error.

apps/ios/LiveVideo/LiveVideo/Views/StreamConfigFlow/EnterStreamNameView.swift
timrozum: Yes this is intended. We get lots of benefits from using a `SwiftUI.Form` on input screens like this. A lot of the styling and layout is done automatically. However if we want to display a big blue button on this screen we can't just add it to the section body because the body will apply extra styling we don't want. Headers and footers don't apply the extra styling and so as a trick we can create a section with empty body and put the button in the footer.

Why not just add this button to the footer of the previous section? Well the sections on this screen are dynamic depending if the user is host or not. And so it would get complex.

It would be nice if there was another way but kind of a small price to pay since we get so many benefits from using `Form` on this screen.

apps/ios/LiveVideo/LiveVideo/Twilio/Sync/SyncStreamDocument.swift
timrozum: Yeah just using the main thread for synchronization. The UI calls the properties on the main thread and the SDK callbacks are on the main thread.

piyushtank (approved):

PR #110, opened 2022-04-07
Make it easier to change Twilio environment

timrozum:
### iOS changes:

1. Allow users to change Twilio environment at runtime similar to the main video app. This is only used by Twilio employees for internal testing.
2. On the sign in screen the user can access settings to change environment.
3. If the environment is not prod, the home screen displays an environment badge so it is obvious to the user what environment they are using.
4. Removed `InAppSettingsKit` and replaced with `SwiftUI.Form`.
5. Removed `SwiftyUserDefaults` and replaced with `@AppStorage` which is a type safe `UserDefaults` property wrapper.

### iOS demo:

https://user-images.githubusercontent.com/1930363/162490809-09adfc2c-090d-4827-8ad0-848440e51c19.mov


wolfspy (approved):
Swift side looks good.

PR #115, opened 2022-04-29
Add iOS chat SDK

timrozum:
We are adding a chat feature to this app so that all users can text chat with each other during the stream.

This PR is focused on basic integration of the Twilio Conversations SDK which will be used for the chat feature. 

The chat UI that was added was just a placeholder so that the SDK could be tested. The placeholder UI will be replaced with a 3rd party chat UI framework most likely in the next PR. The main reason is I think there are a lot of edge cases to handle with the keyboard and also dynamic height text field.

Also updated the player SDK.

### Placeholder chat UI:

Simulator Screen Shot - iPhone 13 Pro - 2022-04-29 at 14 42 56

### DIFF
--- a/apps/ios/LiveVideo/LiveVideo/Twilio/Chat/ChatMessage.swift
+++ b/apps/ios/LiveVideo/LiveVideo/Twilio/Chat/ChatMessage.swift
@@ -5,17 +5,10 @@
 import TwilioConversationsClient
 
 struct ChatMessage: Identifiable {
-    var id: String
-    var author: String
-    var dateCreated: Date
-    var body: String
-    
-    init() {
-        id = UUID().uuidString
-        author = ""
-        dateCreated = Date()
-        body = ""
-    }
+    let id: String
+    let author: String
+    let dateCreated: Date
+    let body: String
 
     init?(message: TCHMessage) {
         guard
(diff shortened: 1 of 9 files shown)


apps/ios/LiveVideo/LiveVideo/Views/Stream/StreamToolbar.swift
```diff
@@ -12,7 +12,7 @@ struct StreamToolbar<Content>: View where Content: View {
     }
     
     var body: some View {
-        HStack(spacing: 10) {
+        HStack(spacing: 0) {
```
timrozum, line 15: Squeeze in a chat button for now. UX will need to figure out how we should add chat.

apps/ios/LiveVideo/LiveVideo/Twilio/Chat/ChatMessage.swift
```diff
@@ -0,0 +1,35 @@
+//
+//  Copyright (C) 2022 Twilio, Inc.
+//
+
+import TwilioConversationsClient
+
+struct ChatMessage: Identifiable {
+    var id: String
```
wolfspy, line 8: Do these need to be vars? It seems incorrect that you would modify something like the `id` later.

wolfspy (changes requested):

apps/ios/LiveVideo/LiveVideo/Twilio/Chat/ChatMessage.swift
timrozum: This was for UI previews but yeah pushed some improvements.

wolfspy (approved):

PR #119, opened 2022-05-06
iOS chat new message badge and bug fixes

timrozum:
1. Made new message badge work.
2. Fixed some bugs that became obvious when testing new message badge.

### Unread dot
Simulator Screen Shot - iPhone 13 Pro - 2022-05-06 at 14 18 32

### DIFF
--- a/apps/ios/LiveVideo/LiveVideo/Views/Chat/ChatMessageList.swift
+++ b/apps/ios/LiveVideo/LiveVideo/Views/Chat/ChatMessageList.swift
@@ -33,27 +33,23 @@ struct ChatMessageList: View {
                         .id("bottom") /// So we can programmatically scroll to this view
                 }
             }
-            .onChange(of: chatManager.hasUnreadMessage) { hasUnreadMessage in
-                guard hasUnreadMessage else {
-                    return
-                }
-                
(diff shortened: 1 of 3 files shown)


apps/ios/LiveVideo/LiveVideo/Views/Chat/ChatMessageList.swift
```diff
@@ -33,27 +33,23 @@ struct ChatMessageList: View {
...
-                /// an Apple bug since people on the web say it used to just work.
-                DispatchQueue.main.async {
-                    /// A convenient way to make the list scroll to the most recent messages at the bottom
-                    chatManager.hasUnreadMessage = true
-                }
+                chatManager.hasUnreadMessage = false
+                
+                /// Must use an extra step so that initial scroll to bottom works. We should be able to perform the scroll right
+                /// here directly but I think there is an Apple bug. People on the web said it used to work but not anymore.
+                isFirstLoadComplete = true
```
timrozum, line 52: I think updating `hasUnreadMessage` from the UI as a result of the UI observing a change to `hasUnreadMessage` was causing some inconsistent behavior. So I added more separation between observing and modifying state. I think the code is actually simpler and seems to fix the issues.

wolfspy (approved):

PR #116, opened 2022-05-03
Fix host creating multiple streams

timmydoza:
**Contributing to Twilio**

> All third party contributors acknowledge that any contributions they provide will be made under the same open source license that the open source project is provided under.

- [ ] I acknowledge that all my contributions will be made under the project's license.

### JIRA link(s):

- [VIDEO-9215](https://issues.corp.twilio.com/browse/VIDEO-9215)

### Description

This PR fixed a bug where the host cannot create more than one stream. It adds disconnect functions for both the SyncProvider and Chat Provider, and calls these functions whenever a participant disconnects from an event. This cleans up the chat and sync states, and there are now no errors when a host tries to create a second stream, because the stale sync and chat instances are no longer present. 

### Before review
* [ ] Added unit tests if necessary
* [ ] Updated affected documentation
* [ ] Verified locally with all effected platforms
* [ ] Manually sanity tested running locally
* [ ] Included screenshot as PR comment (if needed)
* [ ] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

PR #122, opened 2022-05-13
Video 9792 conversations feature flag

olipyskoty:
**Contributing to Twilio**

### JIRA link(s):

- [VIDEO-9792](https://issues.corp.twilio.com/browse/VIDEO-9792)

### Description

This PR adds a feature flag for Twilio Conversations. This will allow developers to easily "turn off" the chat feature for both iOS and web. To disable chat, un-comment `# DISABLE_CHAT=true` in your `.env` file.

**Note**, I closed the previous PR. I had to merge `main` into `feature/audience-chat` so that it included the changes from PR #116. For some reason, when I first did that GitHub was not reflecting those updates in the PR, which would make it confusing when reviewing.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


serverless/functions/create-stream.js
```diff
@@ -293,44 +294,46 @@ module.exports.handler = async (context, event, callback) => {
     return callback(null, response);
   }
 
-  const conversationsClient = client.conversations.services(CONVERSATIONS_SERVICE_SID);
+  if (DISABLE_CHAT !== 'true') {
```
timrozum, line 297: Does this syntax work in JS?
```
if (!DISABLE_CHAT) {
```

timrozum: Or maybe it doesn't because this var is a string? But JS can automatically convert to bool right?

olipyskoty: We can retrieve environment variables in Node.js by accessing the `process.env` object. For some reason, the `process.env` object forces all of its properties to be of the type string. I am pretty sure Twilio Functions appends the `context` object to `process.env`, which is why when chat is disabled, `DISABLE_CHAT` is the string `'true'` instead of the boolean `true`.

I think if we wanted to convert the variable to a boolean I'd have to create a new variable and do something like:
```
let chat_enabled = true;
if (DISABLE_CHAT === 'true') {
     chat_enabled = false
}
```
But I could be wrong, @timmydoza what do you think?

timrozum: Oh interesting, yeah probably best the way it is.

timrozum (approved):
LGTM

serverless/functions/create-stream.js
```diff
@@ -362,6 +367,8 @@ module.exports.handler = async (context, event, callback) => {
       viewers_map: 'viewers',
       raised_hands_map: `raised_hands`,
     },
+    room_sid: room.sid,
```
timmydoza, line 370: Just curious, why are we adding room sid here?

apps/web/src/components/PreJoinScreens/PreJoinScreens.tsx
```diff
@@ -34,7 +34,7 @@ export default function PreJoinScreens() {
       if (appState.hasSpeakerInvite) {
         const { data } = await joinStreamAsSpeaker(appState.participantName, appState.eventName);
         await videoConnect(data.token);
-        chatConnect(data.token);
+        data.chat_enabled && chatConnect(data.token);
```
timmydoza, line 37: Nit - maybe this can be an if statement like the others. Feels funny to use `&&`, even though it works the same.

timmydoza (approved):
Looks good! Just had one style comment...

apps/web/src/components/PreJoinScreens/PreJoinScreens.tsx
olipyskoty: Thanks! Not sure why I handled this differently 🤔

serverless/functions/create-stream.js
olipyskoty: @timrozum mentioned that for iOS it would be best if all participants (host, speaker, viewer) get the room sid the same way for consistency. I figured that this small change wouldn't have any impact for web. What do you think though @timmydoza?

PR #126, opened 2022-06-01
Add basic API documentation for the reference backend

timrozum:
Add API documentation for the reference backend to provide an overview for how the backend is used.

### DIFF
--- a/ReferenceBackendAPI.md
+++ b/ReferenceBackendAPI.md
@@ -152,7 +152,7 @@ Example response:
 
 ### /remove-speaker
 
-This endpoint removes a user from the video room. The endpoint should be called when a host wants to remove a speaker that is misbehaving.
+This endpoint removes a user from the video room. The endpoint should be called when a host wants to remove a speaker that causing shenanigans.
 
 Request parameters:
 
--- a/ReferenceBackendAPI.md
+++ b/ReferenceBackendAPI.md
@@ -10,7 +10,9 @@ Twilio Functions does not utilize HTTP methods so you can just use `POST` for al
 
 All requests must set the `Authorization` header to the correct [passcode](README.md#deploy-the-app-to-twilio). The backend will validate the passcode before processing each request.
 
-If you encounter a `401` error, it may be caused by an incorrect passcode. In addition to using the passcode to authenticate each request, the client also derives part of the backend URL from the passcode. 
+If you encounter a `401` error, it may be caused by either an incorrect passcode, or an expired passcode (passcodes expire after one week). Clients should use the `error.explanation` property from the response to inform users about why their passcode was rejected. 
+
+In addition to using the passcode to authenticate each request, the client also derives part of the backend URL from the passcode. 
 
 ## Errors
 
--- a/ReferenceBackendAPI.md
+++ b/ReferenceBackendAPI.md
@@ -89,7 +89,7 @@ Example response:
 
 ### /join-stream-as-viewer
 
-This endpoint adds the user to the Twilio conversation and adds some user state to some sync objects. The endpoint returns a [Twilio access token](https://www.twilio.com/docs/iam/access-tokens). The token includes grants that allow the client to connect to the live stream, sync objects, and conversation. 
+This endpoint adds the user to the Twilio conversation and adds some user state to some sync objects. The endpoint returns a [Twilio access token](https://www.twilio.com/docs/iam/access-tokens). The token includes grants that allow the client to connect to the Twilio Live Player SDK, Twilio Sync SDK, and Twilio Conversations SDK. 
 The endpoint should be called when a user initially joins the stream as a viewer or when a user transitions from speaker to viewer.
 
 Request parameters:
--- a/ReferenceBackendAPI.md
+++ b/ReferenceBackendAPI.md
@@ -45,7 +45,7 @@ Example response:
 
 ### /create-stream
 
-This endpoint creates a Twilio video room, Twilio Live player streamer, Twilio Live media processor, Twilio sync service, and Twilio conversation. The user is added to the conversation and some user state is updated in sync objects. The endpoint returns a [Twilio access token](https://www.twilio.com/docs/iam/access-tokens). The token includes grants that allow the client to connect to the video room, sync objects, and conversation. The user that creates the stream is the host for that stream.
+This endpoint creates a Twilio video room, Twilio Live player streamer, Twilio Live media processor, Twilio sync service, and Twilio conversation. The user is added to the conversation and some user state is updated in sync objects. The endpoint returns a [Twilio access token](https://www.twilio.com/docs/iam/access-tokens). The token includes grants that allow the client to connect to the Twilio Video SDK, Twilio Sync SDK, and Twilio Conversations SDK. The user that creates the stream is the host for that stream.
 
 Request parameters:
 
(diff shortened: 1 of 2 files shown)


ReferenceBackendAPI.md
```diff
@@ -0,0 +1,193 @@
...
+
+```json
+{
+  "sent": true
+}
+```
+
+### /remove-speaker
+
+This endpoint removes a user from the video room. The endpoint should be called when a host wants to remove a speaker that is misbehaving.
```
timmydoza, line 155:
```suggestion
This endpoint removes a user from the video room. The endpoint should be called when a host wants to remove a speaker that causing shenanigans.
```
Just for fun 🙂 if you want. Also consistent with the `X-Shenanigans: none` header seen on the serverless responses.

ReferenceBackendAPI.md
```diff
@@ -0,0 +1,193 @@
...
+
+```json
+{
+  "verified": true
+}
+```
+
+### /create-stream
+
+This endpoint creates a Twilio video room, Twilio Live player streamer, Twilio Live media processor, Twilio sync service, and Twilio conversation. The user is added to the conversation and some user state is updated in sync objects. The endpoint returns a [Twilio access token](https://www.twilio.com/docs/iam/access-tokens). The token includes grants that allow the client to connect to the video room, sync objects, and conversation. The user that creates the stream is the host for that stream.
```
timmydoza, line 48:
```suggestion
This endpoint creates a Twilio video room, Twilio Live player streamer, Twilio Live media processor, Twilio sync service, and Twilio conversation. The user is added to the conversation and some user state is updated in sync objects. The endpoint returns a [Twilio access token](https://www.twilio.com/docs/iam/access-tokens). The token includes grants that allow the client to connect to the Twilio Video SDK, Twilio Sync SDK, and Twilio Conversations SDK. The user that creates the stream is the host for that stream.
```
What do you think about this? Just to be more specific about how the token is used.

timmydoza (changes requested):
Looks great! I like the look and feel of this over the plugin-rtc docs with the weird tables.

ReferenceBackendAPI.md
timrozum: Oh yeah that is more Twilio.

ReferenceBackendAPI.md
```diff
@@ -0,0 +1,193 @@
...
+
+## HTTP Method
+
+The reference backend is deployed to [Twilio Functions](https://www.twilio.com/docs/runtime/functions), a serverless environment.
+
+Twilio Functions does not utilize HTTP method so you can just use `POST` for all requests.
+
+## Authentication
+
+All requests must set the `Authorization` header to the correct [passcode](README.md#deploy-the-app-to-twilio). The backend will validate the passcode before processing each request.
```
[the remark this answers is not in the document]
timrozum: Yeah we should mention 401 because that can be a confusing dead end.

But regarding the other errors such as expired passcode, the iOS app actually just passes it through for the UI to display just like any other fatal error. And the message that the backend provides for these cases is intuitive. So in other words I think the iOS app works well without any special handling. And I think we recommend that customers completely replace the auth system for a production app. So I imagine a customer wouldn't add custom error handling for the passcode solution. Does the web app handle it differently? I'm thinking no need to get into specific errors?

ReferenceBackendAPI.md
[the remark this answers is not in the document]
timrozum: Sure

ReferenceBackendAPI.md
```diff
@@ -0,0 +1,193 @@
...
+  "error": {
+    "message": "error creating room",
+    "explanation": "Room already exists."
+  }
+}
+```
+
+## Endpoints
+
+### /verify-passcode
```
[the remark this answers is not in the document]
timrozum: See other response

ReferenceBackendAPI.md
```diff
@@ -0,0 +1,195 @@
...
+
+```json
+{
+  "token": "xxxx.yyyyy.zzzzz"
+}
+```
+
+### /join-stream-as-speaker
+
+This endpoint adds the user to the Twilio conversation and adds some user state to the sync objects. The endpoint returns a [Twilio access token](https://www.twilio.com/docs/iam/access-tokens). The token includes grants that allow the client to connect to the video room, sync objects, and conversation. The endpoint should be called when a user initially joins the stream as a speaker or when a user transitions from viewer to speaker.
```
timmydoza, line 71:
```suggestion
This endpoint adds the user to the Twilio conversation and adds some user state to the sync objects. The endpoint returns a [Twilio access token](https://www.twilio.com/docs/iam/access-tokens). The token includes grants that allow the client to connect to the Twilio Video SDK, Twilio Sync SDK, and Twilio Conversations SDK. The endpoint should be called when a user initially joins the stream as a speaker or when a user transitions from viewer to speaker.
```
Just to be consistent with the other section.

ReferenceBackendAPI.md
```diff
@@ -0,0 +1,195 @@
...
+
+```json
+{
+  "token": "xxxx.yyyyy.zzzzz"
+}
+```
+
+### /join-stream-as-viewer
+
+This endpoint adds the user to the Twilio conversation and adds some user state to some sync objects. The endpoint returns a [Twilio access token](https://www.twilio.com/docs/iam/access-tokens). The token includes grants that allow the client to connect to the live stream, sync objects, and conversation.
```
timmydoza, line 92:
```suggestion
This endpoint adds the user to the Twilio conversation and adds some user state to some sync objects. The endpoint returns a [Twilio access token](https://www.twilio.com/docs/iam/access-tokens). The token includes grants that allow the client to connect to the Twilio Live Player SDK, Twilio Sync SDK, and Twilio Conversations SDK. 
```

ReferenceBackendAPI.md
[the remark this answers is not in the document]
timmydoza: That's a good point. I guess we don't need to get into specific errors. But, I still think that it could be useful to mention that there are two different ways to get a 401 here, and that it is also useful (if not necessary) for the UI to consume the error messages that are returned.

ReferenceBackendAPI.md
```diff
@@ -0,0 +1,195 @@
...
+
+The reference backend is deployed to [Twilio Functions](https://www.twilio.com/docs/runtime/functions), a serverless environment.
+
+Twilio Functions does not utilize HTTP methods so you can just use `POST` for all requests.
+
+## Authentication
+
+All requests must set the `Authorization` header to the correct [passcode](README.md#deploy-the-app-to-twilio). The backend will validate the passcode before processing each request.
+
+If you encounter a `401` error, it may be caused by an incorrect passcode. In addition to using the passcode to authenticate each request, the client also derives part of the backend URL from the passcode.
```
timmydoza, line 13:
```suggestion
If you encounter a `401` error, it may be caused by either an incorrect passcode, or an expired passcode (passcodes expire after one week). Clients should use the `error.explanation` property from the response to inform users about why their passcode was rejected. 

In addition to using the passcode to authenticate each request, the client also derives part of the backend URL from the passcode. 
```
What about something like this?

ReferenceBackendAPI.md
```diff
@@ -0,0 +1,195 @@
...
+  }
+}
+```
+
+## Endpoints
+
+### /verify-passcode
+
+Verifies that the passcode in the request header is correct. The client can call this endpoint to validate a passcode the user has entered.
+
```
timmydoza, line 33: It might seem duplicative, but just one sentence here could be nice. Something like:

```
Invalid or expired passcodes will return a `401` error code with an accompanying error object as described above. 
```

timmydoza (changes requested):

timmydoza (approved):
Looks good!

PR #128, opened 2022-06-07
VIDEO-10181 - Android Initial project configuration for Live Video App.

alejo740:
### JIRA link(s):

- [VIDEO-10181](https://issues.corp.twilio.com/browse/VIDEO-10181)

### Description

This is the initial Android project setup for the Android-Live-Video-App

- Set gradle config and dependencies.
- Set initial configuration for Android Manifest.
- Set basic configuration for the initial Fragment and Activity.
- Set configuration to with a base Hilt Module to Inject some dependencies.
- Set initial configuration to provide retrofit instance and interceptor.
- Set App Icon.

### Disclaimers
- Circle CI configuration is still pending.

### Before review
* [ ] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


apps/android/LiveVideoApp/.gitignore
```diff
@@ -0,0 +1,65 @@
+*.iml
```
timrozum, line 1: I did not think to have multiple .gitignore files but I guess it makes quite a lot of sense for a monorepo.

apps/android/LiveVideoApp/.idea/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Default ignored files
```
timrozum, line 1: Oh interesting another .gitignore. So there seems to be some inconsistency. The main .gitignore references sub directories such as `sdk/libs/`. So why wouldn't it also reference `.idea/shelf/` for example? I don't understand why there is inconsistent strategies.

timrozum (commented):
I'm thinking we should add Ardavon to Android PRs also so that we have an Android dev, unless there was a different arrangement made by Tim M.

apps/android/LiveVideoApp/.idea/.gitignore
afalls-twilio: the .gitignore in the .idea folder is created by android studio and can be ignored.

afalls-twilio:
Please delete the .idea folder from the repo

apps/android/LiveVideoApp/.idea/.gitignore
alejo740: Yes, that's right the second .gitignore is auto generated by Android Studio, but yeah, it can be ignored Or I just can remove it.

timrozum: But what about the 3rd .gitignore under the `apps` directory? Is that similar?

timrozum: We talked offline and this .gitignore is also generated automatically. So I'm good with leaving it in.

alejo740:
> Please delete the .idea/ folder from the repo

@afalls-twilio Done!

afalls-twilio (approved):

PR #133, opened 2022-06-14
Connect recording API

timrozum:
1. Configure the `MediaProcessor` to enable recording based on request parameter.
2. Handle recording error in `MediaProcessor` callback.


serverless/constants.js
```diff
@@ -1,7 +1,7 @@
 const syncServicePrefix = 'Live Video '
 
 module.exports = {
-  MEDIA_EXTENSION: 'video-composer-v1',
+  MEDIA_EXTENSION: 'video-composer-v1-test',
```
timrozum, line 4: This will need to be reverted before merging to main branch.

serverless/functions/media-processor-webhook.protected.js
```diff
@@ -0,0 +1,36 @@
...
+exports.handler = async function (context, event, callback) {
+  const common = require(Runtime.getAssets()['/common.js'].path);
+  const { getStreamMapItem, createErrorHandler } = common(context, event, callback);
+
+  const client = context.getTwilioClient();
+
+  const { StatusCallbackEvent, ExtensionContext } = event;
+
+  switch (StatusCallbackEvent) {
+    case 'recording-failed':
```
timrozum, line 12: I wasn't able to fully test this but asked the API engineer if there is a way to force a recording error.

timmydoza: We could maybe test this ourselves by temporarily removing the `.protected` status of this endpoint and sending out own fake webhook event.

timmydoza:
Looks good. 

One thought I had - is maybe we could add a script to help improve the visibility of the finished recording URLs. I put together this quick JS script as an example:

```js
require('dotenv').config();
const client = require('twilio')(process.env.ACCOUNT_SID, process.env.AUTH_TOKEN);
const axios = require('axios');
const axiosClient = axios.create({
  headers: {
    Authorization:
      'Basic ' + Buffer.from(`${process.env.ACCOUNT_SID}:${process.env.AUTH_TOKEN}`, 'utf8').toString('base64'),
    'content-type': 'application/x-www-form-urlencoded;charset=utf-8',
  },
  baseURL:
    client.region === 'dev' || client.region === 'stage'
      ? `https://media.${region}.twilio.com/v1`
      : 'https://media.twilio.com/v1',
});

(async function () {
  const recordings = (await axiosClient('MediaRecordings')).data.media_recordings;
  const rooms = await client.video.rooms.list({ status: 'completed', limit: 100 });

  console.log('\nAvailable Twilio Live Recordings');
  console.log('--------------------------------\n');

  recordings.forEach((recording) => {
    const recordingUrl = recording.links.media;
    const room = rooms.find((room) => room.sid === recording.source_sid);

    console.log(`Event Name: ${room.uniqueName} | Recording URL: ${recordingUrl}`);
  });

  console.log(
    '\nPlease remember to use your Twilio Account SID and Auth Token when accessing the recording URLs in the browser.\n'
  );
})();
```

When run, it looks like this:
image

What do you think? Might also be good to have a PM look at this too.

serverless/functions/media-processor-webhook.protected.js
timrozum: I tested with this:

```
curl -X POST 'twilio-live-interactive-video-XXXX-XXXX-dev.twil.io/media-processor-webhook' \
  -d 'ExtensionContext={"room": {"name": "RMXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"}}' \
  -d 'ErrorMessage=foo' \
  -d 'StatusCallbackEvent=recording-failed'
```

And also modified the webhook to process one of the events that is always fired...I think a "media processor ended" event to make sure some basics worked.

So I think probably the best we can test on our own.

timmydoza (approved):
This looks good! Let's run the script by Donal just to be sure, but I think this does the trick!

serverless/scripts/recordings.js
```diff
@@ -0,0 +1,33 @@
...
+  const rooms = await client.video.rooms.list({ status: 'completed', limit: 100 });
+
+  console.log('\nAvailable Twilio Live Recordings');
+  console.log('--------------------------------\n');
+
+  recordings.forEach((recording) => {
+    const recordingUrl = recording.links.media;
+    const room = rooms.find((room) => room.sid === recording.source_sid);
+
+    console.log(`Event Name: ${room.uniqueName} | Recording URL: ${recordingUrl}`);
```
donaltoomey, line 27: It might be worth printing the "completed" timestamp so that people can identify which recording is which.

timrozum: Sure. I added a new `Date Updated:` column. I figured it was safer to simply call this field the same thing the API called it so that we don't imply things about internal implementation.

timmydoza:
@timrozum Looks good! My only though is about formatting of the script output. At this point it's a lot of info to show on one line. What if we put Event, Date Completed, and Recording URL each on a new line? This is the script that I was playing with:
```
require('dotenv').config();
const client = require('twilio')(process.env.ACCOUNT_SID, process.env.AUTH_TOKEN);
const axios = require('axios');
const axiosClient = axios.create({
  headers: {
    Authorization:
      'Basic ' + Buffer.from(`${process.env.ACCOUNT_SID}:${process.env.AUTH_TOKEN}`, 'utf8').toString('base64'),
    'content-type': 'application/x-www-form-urlencoded;charset=utf-8',
  },
  baseURL:
    client.region === 'dev' || client.region === 'stage'
      ? `https://media.${region}.twilio.com/v1`
      : 'https://media.twilio.com/v1',
});

(async function () {
  const recordings = (await axiosClient('MediaRecordings')).data.media_recordings;
  const rooms = await client.video.rooms.list({ status: 'completed', limit: 100 });

  console.log('\nAvailable Twilio Live Recordings');
  console.log('--------------------------------\n');

  recordings
    .filter((recording) => recording.status == 'COMPLETED')
    .forEach((recording) => {
      const recordingUrl = recording.links.media;
      const dateUpdated = Date(recording.date_updated).toLocaleString();

      // If we don't find the room in the first page of rooms, use SID instead of name
      const room = rooms.find((room) => room.sid === recording.source_sid)?.uniqueName ?? recording.source_sid;

      console.log(`Event Name: ${room}\nDate updated: ${dateUpdated}\nRecording URL: ${recordingUrl}\n`);
    });

  console.log(
    '\nPlease remember to use your Twilio Account SID and Auth Token when accessing the recording URLs in the browser.\n'
  );
})();
```
Also, were we going to go with Recording Completed instead of Recording Updated?

PR #135, opened 2022-06-22
Feature/video 10184/api calls to sign in flow and secure persistence session

alejo740:
**Contributing to Twilio**

## Pull Request Details

- Commit 1:
  - Implemented Secure LocalStorage with Encrypted SharedPreferences.
  - Implemented RemoteStorage to run REST API calls.
  - Implemented TwilioLiveService interface connected with Retrofit instance.
  - Implemented TwilioLiveRepository responsible to manage the data access to LocalStorage or RemoteStorage.
  - Implemented AuthenticatorManager in order to manage all task related to the passcode.
  - Added all the providers to MainModule, in order to let Hilt manage all the Dependencies injections.

- Commit 2:
  - Implemented passcode verification connecting with Rest API
  - Implemented logic in the Interceptor to customize the base URL based on passcode.
  - Added validation to check stored credentials then route to Home Screen.
  - Added Util methods inside of AuthenticatorManager to manipulate the passcode.
  - Implemented generic error case when the passcode is not valid.
  - Added logic in the RemoteStorage to parse the error body when is needed.

### JIRA link(s):
- [VIDEO-10183](https://issues.corp.twilio.com/browse/VIDEO-10183)
- [VIDEO-10184](https://issues.corp.twilio.com/browse/VIDEO-10184)

### Description

In this PR the app is connected with the Local Storage to store the passcode and also to validate and keep the session on when the user already signed-in, then is redirected to the Home Screen.
Also, is implemented the first API call to `/verify-passcode` and connected to the UI to run the proper actions.
Implemented the first generic error flow, if the passcode is not valid.

### Demo
https://user-images.githubusercontent.com/929504/175079959-1c5f81f8-95ee-4993-8d1b-31dadddb0964.mp4

### Before review
* [ ] Verified locally with all effected platforms
* [ ] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


afalls-twilio:
You should add some unit test cases to verify that interaction with the rest api is working.

afalls-twilio (approved):
As long as you add unit test cases to verify the REST interactions and password management with a later PR, I'm good.

alejo740:
> I suppose you will be adding test cases with a different task?

@afalls-twilio yes, I will be adding Unit Test cases the coming sprint. This is the task: [VIDEO-10441](https://issues.corp.twilio.com/browse/VIDEO-10441)

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/repository/model/GenericResponse.kt
```diff
@@ -0,0 +1,42 @@
...
+    val isSuccess: Boolean? = null,
+
+    @SerializedName("sent")
+    val isSent: Boolean? = null,
+
+    @SerializedName("removed")
+    val isRemoved: Boolean? = null,
+
+    @SerializedName("token")
+    val token: String? = null
```
timrozum, line 20: I think it's great to have a reusable `ErrorResponse` and maybe even a way to reuse the token response since it is the same for 3 endpoints.

But I wonder about having a single response that works for any endpoint in the app like this. It seems like it is less type safe because everything has to be optional. And while this approach may work in this reference app because the responses are simple and there are not that many requests, I wonder if this is standard best practice.

What are the benefits of this approach? Is this a popular Android convention? Or am I misunderstanding something?

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/repository/model/GenericResponse.kt
```diff
@@ -0,0 +1,42 @@
...
+
+) : BaseResponse()
+
+data class ErrorResponse(
+
+    @SerializedName("message")
+    val message: String? = null,
+
+    @SerializedName("explanation")
+    val explanation: String? = null
```
timrozum, line 30: The backend API is such that these will never be null. Would you still make them optional in Android?

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/view/SignInPasscodeFragment.kt
```diff
@@ -29,6 +29,6 @@ class SignInPasscodeFragment : SignInBaseFragment() {
     }
 
     private fun registerSignInPasscodeStringObserver() {
-        viewModel.passcode.observe(viewLifecycleOwner) { viewModel.enablePasscodeContinue(it.isNotEmpty()) }
+        viewModel.passcode.observe(viewLifecycleOwner) { viewModel.enablePasscodeContinue(it.isNotEmpty() && it.length >= 8) }
```
timrozum, line 32: Looks like you were able to figure out a lot of the passcode details. Good job given I don't think the internals are documented. A few things:

1. I assume `length >= 8` makes `isNotEmpty` redundant
2. I think the [min length for the passcode is actually 14](https://github.com/twilio/twilio-live-interactive-video/blob/ac978a80113d3a4dcfe210f7f58014c34848d78a/apps/ios/LiveVideo/LiveVideo/Managers/Auth/PasscodeComponents.swift#L14-L16) unless I'm missing something
3. If we are going to verify specific length it may be good to move the min length constant to some centralized file that contains all passcode information.
4. In this situation I think iOS just verifies the string is not empty and that's it. That solution would probably be ok also.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/AuthenticatorManager.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    fun getPasscodeURL(): String = localStorage.getStringData(URL_PASSCODE_KEY) ?: ""
+
+    fun isPasscodeValid(): Boolean = getPasscode().isNotEmpty()
+
+    fun extractPasscodeUrl(passcode: String): String {
+        var passcodeUrl = passcode.substring(passcode.length - 8)
+        passcodeUrl =
+            "${passcodeUrl.substring(0, 4)}-${passcodeUrl.substring(4, passcodeUrl.length)}"
+        return passcodeUrl.trim()
```
timrozum, line 25: There may be an issue with line 22. We designed the passcode so that the serverless ID on the end can be any length >= 4. And so I think if you start parsing the string from the end instead of the beginning there may be issues, if not now in the future.

You might take a look at https://github.com/twilio/twilio-live-interactive-video/blob/main/apps/ios/LiveVideo/LiveVideo/Managers/Auth/PasscodeComponents.swift to see if you like any ideas in this file. It encapsulates all of the passcode decoding in a separate file. 

Also the passcode decoding might be a good way to add some unit tests. It should be easy to test and high value. These are the tests I made for iOS: https://github.com/twilio/twilio-live-interactive-video/blob/main/apps/ios/LiveVideo/LiveVideoTests/PasscodeComponentsSpec.swift

timrozum (changes requested):

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/repository/TwilioLiveRepository.kt
```diff
@@ -0,0 +1,23 @@
+package com.twilio.livevideo.app.repository
+
+import com.twilio.livevideo.app.manager.AuthenticatorManager
+import com.twilio.livevideo.app.repository.datasource.remote.RemoteStorage
+import com.twilio.livevideo.app.repository.model.GenericResponse
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+
+class TwilioLiveRepository(
```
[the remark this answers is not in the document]
alejo740: Ok, similar case, the repository concept, is to interact with specific data coming for specific `Features`. In this case I am thinking just to keep only one repository layer for the entire app, as long as this app is only interacting with a baseUrl and endpoints of `LiveVideoApp`. So, probably renaming it, would be good... something like: `LiveVideoRepository`

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/repository/datasource/remote/TwilioLiveService.kt
```diff
@@ -0,0 +1,14 @@
+package com.twilio.livevideo.app.repository.datasource.remote
+
+import com.twilio.livevideo.app.network.LiveAppInterceptor
+import com.twilio.livevideo.app.repository.model.GenericResponse
+import retrofit2.Response
+import retrofit2.http.GET
+import retrofit2.http.Header
+
+interface TwilioLiveService {
```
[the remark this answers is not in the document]
alejo740: Oh yeah, thats good, renaming to: `LiveVideoAPIService`

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/repository/model/GenericResponse.kt
alejo740: Oh no, actually I tested the api calls using Postman, and... if the response is success, then the error body is coming `null`.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/view/SignInPasscodeFragment.kt
alejo740: Good catch Tim!

1. You right, I will remove `isNotEmpty` condition.
2. Yeah, I know the full passcode is 14 digits, but only the last 8 digits and the actual digits used to map into the URL. So, thats the reason of that `8` limit.
3. Sure, I will move this and some other numbers to be constants. 
4. Yes, iOS only verify if it's not empty, but in this case, I implemented a logic to extract the last 8 digits, but if the entered passcode is less than 8 digits, then it cause error. So, i decided to place that logic/validation in the View/UI.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/network/LiveAppInterceptor.kt
```diff
@@ -1,19 +1,39 @@
...
+import com.twilio.livevideo.app.manager.AuthenticatorManager
+import okhttp3.HttpUrl
 import okhttp3.Interceptor
 import okhttp3.Response
 import javax.inject.Inject
-import javax.inject.Singleton
 
-@Singleton
-class LiveAppInterceptor @Inject constructor() : Interceptor {
+class LiveAppInterceptor @Inject constructor(private val authenticatorManager: AuthenticatorManager) :
```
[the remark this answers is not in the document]
timrozum: I think `RequestInterceptor` works well for this app.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/repository/TwilioLiveRepository.kt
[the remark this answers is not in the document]
timrozum: Yeah that works.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/repository/model/GenericResponse.kt
timrozum: Yeah I understand. I think it's probably best to map each response, so it's more conventional for others looking at the code.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/view/SignInPasscodeFragment.kt
timrozum: Oh I understand why you were using 8. Yeah named constants would help make that more self documenting.

But I think there is still an issue that could happen since the component of the passcode can actually be variable length. I can explain offline.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/AuthenticatorManager.kt
alejo740: As per our conversation, I changed this logic to process the passcode.
Regarding the Unit test. Yes, I will be adding some unit test on a different ticket. Thanks for your tests on iOS, I think I will use that as a reference to make it on Android.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/repository/model/GenericResponse.kt
timrozum: Now that the responses are separated I think a lot of the fields don't need to be optional anymore. For example:
```
val isVerified: Boolean? = null
```

timrozum (approved):
We talked a lot offline about error handling. I think for this PR we can just keep it simple for now. 

As the app evolves we can continue to improve the architecture so that the app displays the most useful error information to the user. 

For example if there is a JSON parsing error from an API response, we should be able to display the parsing error details in the app.

We can evolve this over time.

timrozum (approved):

PR #136, opened 2022-06-27
Video 10105 finalize audience chat

olipyskoty:
**Contributing to Twilio**

### JIRA link(s):

- [VIDEO-10105](https://issues.corp.twilio.com/browse/VIDEO-10105)

### Description

This PR has the chat window open by default for desktop. It also fixes a bug where the chat button was not being displayed for viewers even when Chat was enabled 😟. Finally, it cleans up the bottom menu bar for mobile (for all types of participants)

**Speaker**
IMG_8137

**Viewer**
IMG_8136

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


apps/web/src/App.tsx
```diff
@@ -50,7 +52,7 @@ export default function App() {
           <RecordingNotifications />
           <MobileTopMenuBar />
           <Room />
-          <MenuBar />
+          {isMobile ? <MobileBottomMenuBar /> : <MenuBar />}
```
timmydoza, line 55: I think we should try to just use `<MenuBar />` only, but if we had `<MobileBottomMenuBar />` too then we should use this to switch between the two:
```
  const isMobile = useMediaQuery((theme: Theme) => theme.breakpoints.down('sm'));
```
`isMobile` from `utils` just looks at the user agent. For styling on mobile, we should use media queries.

apps/web/src/components/Buttons/LeaveEventButton/LeaveEventButton.tsx
```diff
@@ -54,7 +54,7 @@ export default function LeaveEventButton(props: { buttonClassName?: string }) {
   return (
     <>
       <Button onClick={() => setMenuOpen(isOpen => !isOpen)} ref={anchorRef} className={classes.button}>
-        Leave Event
+        {!props.excludeLabel && 'Leave Event'}
```
timmydoza, line 57: Can this work instead in all buttons?
```
<Hidden smDown>
  Leave Event
</Hidden>
```

apps/web/src/components/MobileBottomMenuBar/MobileBottomMenuBar.tsx
```diff
@@ -0,0 +1,95 @@
...
+);
+
+export default function MenuBar() {
+  const classes = useStyles();
+  const { isSharingScreen, toggleScreenShare } = useVideoContext();
+  const roomState = useRoomState();
+  const isReconnecting = roomState === 'reconnecting';
+  const { appState } = useAppState();
+
+  return (
```
timmydoza, line 72: Much of this seems very similar to the MenuBar component. I wonder if we truly need a new MobileBottomMenuBar component. Can we use `<Hidden />` in MenuBar to get it to work on mobile?

apps/web/src/components/Player/PlayerMenuBar/PlayerMenuBar.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { useCallback, useRef, useState } from 'react';
 import { createStyles, makeStyles, Theme } from '@material-ui/core/styles';
 import { Typography, Grid, Button } from '@material-ui/core';
+import { isMobile } from '../../../utils';
```
timmydoza, line 4: Will `<Hidden>` work instead?

Also we should not use this `isMobile` to adjust styles. Using the userAgent to adjust styles on mobile is not responsive - the UI cannot adjust itself in response to changes in browser width.

timmydoza (changes requested):

apps/web/src/components/MenuBar/MenuBar.tsx
```diff
@@ -31,6 +31,23 @@ const useStyles = makeStyles((theme: Theme) =>
       [theme.breakpoints.down('sm')]: {
         height: `${theme.mobileFooterHeight}px`,
         padding: 0,
+        '& .MuiButton-textLabel': {
+          display: 'none',
+        },
+        '& .MuiButton-mobileBackground': {
+          marginRight: '1em',
```
timmydoza, line 38:
```suggestion
          margin: '0 0.5em',
```
This will cause the buttons to be perfectly centered. With marginRight, it's a little lopsided. There's a little more white space on the right than on the left:

image

timmydoza (changes requested):

timmydoza (approved):

PR #139, opened 2022-06-29
Small readme change

timrozum:
Make it clearer that the host can enable recording in the app UI as opposed to the API.


guifreballester (approved):
looks great!

PR #140, opened 2022-06-29
Add support for media recordings

timrozum:
1. Add recording support to the web app, iOS app, and reference backend. The event host can enable recording in the app UI when they create the event.
2. Display alert when a user joins a stream to explain that recording is enabled.
3. Display recording indicator while stream is in progress to indicate recording is ongoing.
4. Use `npm run recordings` to view recordings.
5. Refactor error handling in the reference backend to increase code reuse and make the code easier to follow.


serverless/constants.js
```diff
@@ -1,7 +1,7 @@
 const syncServicePrefix = 'Live Video '
 
 module.exports = {
-  MEDIA_EXTENSION: 'video-composer-v1',
+  MEDIA_EXTENSION: 'video-composer-v1-test',
```
timrozum, line 4: We should change this to `video-composer-v2` eventually.

PR #150, opened 2022-07-21
Feature/video 10585/android create stream and room sdk integration

alejo740:
**Contributing to Twilio**

### JIRA link(s):
- [VIDEO-10585](https://issues.corp.twilio.com/browse/VIDEO-10585)

### Description
- Added conditions for Host role
- Implemented API calls to create-stream and delete-stream
- Integrated Room SDK connecting to a new Streaming using the accessToken
- Implemented logic to handle one Local Participant and multiple Remote Participants.
- Added permissions support for AUDIO_RECORD and CAMERA
- Connected the UI only to display the local video track for demo purposes

## Demo

https://user-images.githubusercontent.com/929504/180112089-c2065ba3-c05b-46bd-8cc0-c7fb71b6e20e.mp4

### DIFF
--- a/apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/CameraCapturerCompat.kt
+++ b/apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/CameraCapturerCompat.kt
@@ -1,0 +1,135 @@
+package com.twilio.livevideo.app.manager.room
+
+import android.content.Context
+import android.graphics.ImageFormat
+import android.hardware.camera2.CameraCharacteristics
+import android.hardware.camera2.CameraManager
+import android.hardware.camera2.CameraMetadata
+import android.os.Build
+import com.twilio.video.Camera2Capturer
+import com.twilio.video.CameraCapturer
+import com.twilio.video.VideoCapturer
+import timber.log.Timber
+import tvi.webrtc.Camera1Enumerator
+import tvi.webrtc.Camera2Enumerator
+import tvi.webrtc.CameraEnumerator
+import tvi.webrtc.CapturerObserver
+import tvi.webrtc.SurfaceTextureHelper
+
+class CameraCapturerCompat(
+    private val frontCameraId: String?,
+    private val backCameraId: String?,
+    private val cameraCapturer: CameraCapturer? = null,
+    private val camera2Capturer: Camera2Capturer? = null
+) : VideoCapturer {
+
+    val cameraId: String
+        get() = cameraCapturer?.cameraId
+                ?: camera2Capturer?.cameraId
+                ?: throw IllegalStateException("At least one camera capturer must not be null")
+
+    override fun initialize(
+        surfaceTextureHelper: SurfaceTextureHelper,
+        context: Context,
+        capturerObserver: CapturerObserver
+    ) {
+        cameraCapturer?.initialize(surfaceTextureHelper, context, capturerObserver)
+                ?: camera2Capturer?.initialize(surfaceTextureHelper, context, capturerObserver)
+    }
+
+    override fun startCapture(width: Int, height: Int, framerate: Int) {
+        cameraCapturer?.startCapture(width, height, framerate) ?: camera2Capturer?.startCapture(width, height, framerate)
+    }
+
+    override fun stopCapture() {
+        cameraCapturer?.stopCapture() ?: camera2Capturer?.stopCapture()
+    }
+
+    override fun isScreencast() = cameraCapturer?.isScreencast ?: camera2Capturer?.isScreencast ?: false
+
+    fun switchCamera() {
+        if (frontCameraId != null && backCameraId != null) {
+            val newCameraId = if (cameraId == frontCameraId) backCameraId else frontCameraId
+            cameraCapturer?.switchCamera(newCameraId)
+            camera2Capturer?.switchCamera(newCameraId)
+        } else Timber.w("Front and back cameras need to both be available in order to switch between them")
+    }
+
+    companion object {
+        fun newInstance(context: Context): CameraCapturerCompat? {
+            return if (Camera2Capturer.isSupported(context)) {
+                Camera2Enumerator(context).getFrontAndBackCameraIds(context)?.let { cameraIds ->
+                    val cameraCapturer = Camera2Capturer(context, cameraIds.first
+                            ?: cameraIds.second ?: "")
+                    CameraCapturerCompat(cameraIds.first, cameraIds.second, camera2Capturer = cameraCapturer)
+                }
+            } else {
+                Camera1Enumerator().getFrontAndBackCameraIds(context, isCamera2 = false)?.let { cameraIds ->
+                    val cameraCapturer = CameraCapturer(context, cameraIds.first ?: cameraIds.second
+                    ?: "", getCameraListener())
+                    CameraCapturerCompat(cameraIds.first, cameraIds.second, cameraCapturer = cameraCapturer)
+                }
+            }
+        }
+
+        private fun getCameraListener() = object : CameraCapturer.Listener {
+            override fun onFirstFrameAvailable() { }
+
+            override fun onCameraSwitched(newCameraId: String) {}
+
+            override fun onError(errorCode: Int) {}
+        }
+
+        private fun CameraEnumerator.getFrontAndBackCameraIds(context: Context, isCamera2: Boolean = true): Pair<String?, String?>? {
+            val cameraIds = deviceNames.find { isFrontFacing(it) && isCameraIdSupported(isCamera2, context, it) } to
+                    deviceNames.find { isBackFacing(it) && isCameraIdSupported(isCamera2, context, it) }
+            return if (isAtLeastOneCameraAvailable(cameraIds.first, cameraIds.second)) cameraIds
+            else {
+                Timber.w("No cameras are available on this device")
+                null
+            }
+        }
+
+        private fun isCameraIdSupported(isCamera2: Boolean, context: Context, cameraId: String) =
+                if (isCamera2) isCameraIdSupported(context, cameraId) else true
+
+        private fun isCameraIdSupported(context: Context, cameraId: String): Boolean {
+            val cameraManager = context.getSystemService(Context.CAMERA_SERVICE) as CameraManager
+            var isMonoChromeSupported = false
+            var isPrivateImageFormatSupported = false
+            val cameraCharacteristics: CameraCharacteristics = try {
+                cameraManager.getCameraCharacteristics(cameraId)
+            } catch (e: Exception) {
+                Timber.e(e)
+                return false
+            }
+            /*
+             * This is a temporary work around for a RuntimeException that occurs on devices which contain cameras
+             * that do not support ImageFormat.PRIVATE output formats. A long term fix is currently in development.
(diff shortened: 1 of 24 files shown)


afalls-twilio (approved):

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/CameraCapturerCompat.kt
```diff
@@ -0,0 +1,135 @@
...
+import tvi.webrtc.Camera2Enumerator
+import tvi.webrtc.CameraEnumerator
+import tvi.webrtc.CapturerObserver
+import tvi.webrtc.SurfaceTextureHelper
+
+class CameraCapturerCompat(
+    private val frontCameraId: String?,
+    private val backCameraId: String?,
+    private val cameraCapturer: CameraCapturer? = null,
+    private val camera2Capturer: Camera2Capturer? = null
```
timrozum, line 23: What does `camera2Capturer` do?

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/CameraCapturerCompat.kt
```diff
@@ -0,0 +1,135 @@
...
+
+    override fun startCapture(width: Int, height: Int, framerate: Int) {
+        cameraCapturer?.startCapture(width, height, framerate) ?: camera2Capturer?.startCapture(width, height, framerate)
+    }
+
+    override fun stopCapture() {
+        cameraCapturer?.stopCapture() ?: camera2Capturer?.stopCapture()
+    }
+
+    override fun isScreencast() = cameraCapturer?.isScreencast ?: camera2Capturer?.isScreencast ?: false
```
timrozum, line 48: What is `isScreencast` for?

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/LocalParticipantWrapper.kt
```diff
@@ -0,0 +1,182 @@
...
+
+    private fun removeAudioTrack() {
+        localAudioTrack?.let { localAudioTrack ->
+            unPublishTrack(localAudioTrack)
+            localAudioTrack.release()
+            this.localAudioTrack = null
+        }
+    }
+
+    override fun onNetworkQualityLevelChanged(localParticipant: LocalParticipant, networkQualityLevel: NetworkQualityLevel) {
```
timrozum, line 129: The iOS app does not use this and so I would recommend removing it for Android.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/LocalParticipantWrapper.kt
```diff
@@ -0,0 +1,182 @@
...
+        this.networkQualityLevel = networkQualityLevel
+    }
+
+    override fun onAudioTrackPublished(
+        localParticipant: LocalParticipant,
+        localAudioTrackPublication: LocalAudioTrackPublication
+    ) {
+        //TODO("Not yet implemented")
```
timrozum, line 137: I think for local participant this app only needs to handle the error callbacks. And so this happy path callback for example can probably just be removed.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/RoomManager.kt
```diff
@@ -0,0 +1,165 @@
...
+            val connectOptions: ConnectOptionsBuilder = {
+                this.roomName(roomName)
+                this.audioTracks(listOf(localParticipantWrapper.localAudioTrack))
+                this.videoTracks(listOf(localParticipantWrapper.videoTrack))
+                this.enableDominantSpeaker(true)
+                this.bandwidthProfile(createBandwidthProfileOptions {
+                    this.mode(BandwidthProfileMode.GRID)
+                    this.dominantSpeakerPriority(TrackPriority.HIGH)
+                })
+                preferVideoCodecs(listOf(Vp8Codec()))
```
timrozum, line 68: I'm thinking we want VP8 simulcast. Is that not what the Android video app uses by default?

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/RoomManager.kt
```diff
@@ -0,0 +1,165 @@
...
+
+    override fun onDominantSpeakerChanged(room: Room, remoteParticipant: RemoteParticipant?) {
+        super.onDominantSpeakerChanged(room, remoteParticipant)
+        participants?.first { it.wrapper.isDominantSpeaker }?.apply { this.wrapper.isDominantSpeaker = false }
+        participants?.first { it.wrapper.identity == remoteParticipant?.identity }?.apply {
+            this.wrapper.isDominantSpeaker = true
+        }
+    }
+
+    private fun RemoteParticipant.isVideoComposer(): Boolean =
```
timrozum, line 163: Interesting syntax!

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/CameraCapturerCompat.kt
```diff
@@ -0,0 +1,135 @@
...
+import com.twilio.video.CameraCapturer
+import com.twilio.video.VideoCapturer
+import timber.log.Timber
+import tvi.webrtc.Camera1Enumerator
+import tvi.webrtc.Camera2Enumerator
+import tvi.webrtc.CameraEnumerator
+import tvi.webrtc.CapturerObserver
+import tvi.webrtc.SurfaceTextureHelper
+
+class CameraCapturerCompat(
```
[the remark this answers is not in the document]
alejo740: `Compat` stands for `Compatibility`. This is the standard way in Android framework to call those classes to make a feature compatible with old versions. 

In this case, this is a helper class to determine what `CameraCapture` instance/version to use on demand, depending on the Android API version. 

So, in Android there are two APIs for Camera Capturer: `CameraCapturer` and `Camera2Capturer`. 
- `CameraCapture`: is a native android class for Android API versions older than `LOLLIPOP` (Android SDK 21)
- `Camera2Capture`: is a native android class for the new versions, compatible with Android API `LOLLIPOP` and newer.

So, this class will decided what version Camera Capture API to choose on demand. I am reusing this entire class from `Android Video Room App`.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/CameraCapturerCompat.kt
alejo740: I think, I replied this on the previous comment: https://github.com/twilio/twilio-live-interactive-video/pull/150#discussion_r927078658

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/CameraCapturerCompat.kt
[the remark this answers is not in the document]
timrozum: I understand now thanks.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/CameraCapturerCompat.kt
alejo740: This is just a method to override the behavior when in case, the capturer is like a screen casting service to external devices. You may have seen a feature to cast the screen to other Android/Google screens like TVs, but for now, this is not being used. However it's required to override from the interface `VideoCapturer` coming from `webrtc` libraries.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/LocalParticipantWrapper.kt
alejo740: Ohhh ok, I implemented this very easy ways, because I see in the `AndroidVideoRoomApp` it was implemented, so I thought would it be nice to have this, just in case for future.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/LocalParticipantWrapper.kt
alejo740: Yes, you right, I haven't figure it out, what methods are actually that we are going to use from this listener. However, in `Kotlin` and `Java`, as we implement an `Interface`, we should override/implement ALL the methods from that interface. In this case, we are forced to implement all the non-optional/non-default methods from the interface `LocalParticipant.Listener`

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/RoomManager.kt
alejo740: Yes, according to `AndroidVideoRoomApp` this is the default video codec(without simulcast). However, if we want to use `VP8 simulcast`, when I just need to pass true as parameter. Something like: `Vp8Codec(true)`. Let me know.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/RoomManager.kt
alejo740: 😄 yes, this is the way to extend functions for a specific already defined class.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/LocalParticipantWrapper.kt
timrozum: Makes sense and seems better actually. That is the same with pure Swift, but because the SDK is Objective-C each callback is optional.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/LocalParticipantWrapper.kt
timrozum: I would recommend only including code that we really need in order to minimize maintenance and make the code as simple as possible to follow for customers.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/RoomManager.kt
alejo740: Done!

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/LocalParticipantWrapper.kt
alejo740: Done!

timrozum (approved):

PR #154, opened 2022-07-29
task/VIDEO-10785/ Refactor_RemoteStorage_to_parse_error_body_from_Interceptor

alejo740:
### JIRA link(s):
- [VIDEO-10785](https://issues.corp.twilio.com/browse/VIDEO-10785)

### Description

Refactor for RemoteStorage and LiveVideoRequestInterceptor in order to parse the error body from the interceptor.

However, all the parsing process to be moved to the `Interceptor` was not completely achieved, because `OkHttp` library that works underneath of `Retrofit` only allow internal parsing for Success responses, but not for Error responses. The reason behind of this, is to give the flexibility to `convert/parse` the error `json response` to any model class. 

So, the only way to achieve it was with a `Hack` changing the error `response code` to any success code (in this case: `299`) in the response from the interceptor. This, to pretend the internal parsing process to “simulate” the response was `Success` then validate it in the `RemoteStorage` class when the response has returned.

### DIFF
--- a/apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/network/LiveVideoRequestInterceptor.kt
+++ b/apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/network/LiveVideoRequestInterceptor.kt
@@ -1,14 +1,19 @@
 package com.twilio.livevideo.app.network
 
+import com.google.gson.Gson
 import com.twilio.livevideo.app.manager.AuthenticatorManager
 import com.twilio.livevideo.app.util.PasscodeUtil
 import okhttp3.HttpUrl
 import okhttp3.Interceptor
 import okhttp3.Response
+import okhttp3.ResponseBody
+import timber.log.Timber
 import javax.inject.Inject
 
 class LiveVideoRequestInterceptor @Inject constructor(private val authenticatorManager: AuthenticatorManager) :
     Interceptor {
+
+    private val gson = Gson()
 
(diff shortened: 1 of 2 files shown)


timrozum (commented):
I think I will have to defer to @afalls-twilio for this PR since it is getting deep into Android details. 

However, it does seems like there should be a way to make the error handling code more reusable.

alejo740:
> I think I will have to defer to @afalls-twilio for this PR since it is getting deep into Android details.
> 
> However, it does seems like there should be a way to make the error handling code more reusable.

Yes!... I think the current implementation is a good approach to make the error handling reusable. Using the method `processResponse`

afalls-twilio (approved):
Good job on the error interceptor

timrozum:
> > I think I will have to defer to @afalls-twilio for this PR since it is getting deep into Android details.
> > However, it does seems like there should be a way to make the error handling code more reusable.
> 
> Yes!... I think the current implementation is a good approach to make the error handling reusable. Using the method `processResponse`

I think my comment may have been confusing. To be clear, I think calling `processResponse` for each request is duplicate code and it seems like there should be a good way to reuse more code so that each request does not have to make a separate call to `processResponse`. 

I do not have any specific solutions however since this is getting into Android details.

alejo740:
I close this PR due to this approach is not a good practice. So, instead a new PR is created with a new approach: https://github.com/twilio/twilio-live-interactive-video/pull/156

PR #161, opened 2022-08-11
task/VIDEO-10999 Optimize grid mode

alejo740:
### JIRA link(s):
- [VIDEO-10999](https://issues.corp.twilio.com/browse/VIDEO-10999)

### Description

- Optimize the grid rendering, to avoid remove and add existing participants UI thus avoid flickering.
- Created `GridManager.kt` in order to manage all the UI GridMode operations.
- Implemented Switch View when a Dominant Speaker changed.

### Demo
For this demo purposes, the Off-Screen Participants Limit was changed to `4`.

https://user-images.githubusercontent.com/929504/184175172-c292098c-2a01-4a79-b600-42b60b01ef18.mp4

### DIFF
--- a/apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/GridManager.kt
+++ b/apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/GridManager.kt
@@ -1,6 +1,5 @@
 package com.twilio.livevideo.app.manager
 
-import android.content.Context
 import android.view.LayoutInflater
 import android.view.ViewGroup
 import android.widget.GridLayout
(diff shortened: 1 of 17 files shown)


apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/GridManager.kt
```diff
@@ -0,0 +1,153 @@
...
+        fun switchParticipants(list: MutableList<ParticipantStream>, participant: ParticipantStream): Boolean {
+            list.indexOf(participant).also { index ->
+                if (index + 1 > GRID_MAX_PARTICIPANTS_LIMIT) {
+                    //Take the first N limit participants and skip the first element which is the local participant.
+                    list.take(GRID_MAX_PARTICIPANTS_LIMIT).drop(1).minByOrNull {
+                        it.dominantSpeakerStartTime
+                    }?.also { oldestDominantSpeaker ->
+                        val oldestDominantSpeakerIndex = list.indexOf(oldestDominantSpeaker)
+                        list[oldestDominantSpeakerIndex] = participant
+                        list[index] = oldestDominantSpeaker
```
timrozum, line 145: Looks like this code is swapping participants. iOS and web apps behave differently. When the oldest dominant speaker is removed from the grid, they are always moved to the start of the offscreen list. This way if someone on the grid disconnects later, the most recent dominant speaker off screen will be moved offscreen.

See how this is a 4 step process on iOS: https://github.com/twilio/twilio-live-interactive-video/blob/eba1c8a78235cc22a90cf2bfd3780d6b267dfbf3/apps/ios/LiveVideo/LiveVideo/Views/Stream/SpeakerGridViewModel.swift#L110-L113

timrozum: Also it seems like `GridManager` should not modify internal state of `RoomManager`. It gets to be difficult to follow all the places the state is being modified.

It seems like the grid state is split between `GridManager` and `RoomManager`. I would recommend having `GridManager` own state for all participants regardless if they are visible or not on the grid. 

Eventually you will be implementing speaker layout which is a completely different layout. The User will be able to switch between grid and speaker layouts. It is important for each layout to manage their own state so that one layout has no impact on another layout.

timrozum (changes requested):
Looks like a good solution for the UI view.

I think we just need to have the grid model own all of the grid state and have a clearer boundary between the grid and `RoomManager`.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/GridManager.kt
alejo740: Done!

timrozum (approved):
Nice!

PR #162, opened 2022-08-26
task/VIDEO-11001 android transition move speaker as viewer

alejo740:
### JIRA link(s):
- [VIDEO-11001](https://issues.corp.twilio.com/browse/VIDEO-11001)

### Description
- Implemented transition Speaker to Viewer by themself
- Implemented transition Speaker to Viewer by host
- Implemented API call to removeSpeaker
- Implemented UI popup menu for speaker bottom controls
- Implemented UI popup menu for each remote participants to allow Host moves speakers as viewers.

## Demo
- ### Host sends speaker as viewer
https://user-images.githubusercontent.com/929504/186794212-edb65f14-9bc9-4c33-b720-8bebb22756bd.mp4

- ### Speaker is sent as viewer by host
https://user-images.githubusercontent.com/929504/186981625-849224f3-4ba7-4949-8109-bcc365f690f9.mp4

- ### Speaker choose the move as viewer by themselves 
https://user-images.githubusercontent.com/929504/186982486-72b26f21-e6f7-4cfe-878f-2340dbef616a.mp4

### DIFF
--- a/apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/GridManager.kt
+++ b/apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/GridManager.kt
@@ -1,21 +1,50 @@
 package com.twilio.livevideo.app.manager
 
 import android.view.LayoutInflater
+import android.view.View
 import android.view.ViewGroup
 import android.widget.GridLayout
+import androidx.appcompat.widget.PopupMenu
 import androidx.core.view.doOnAttach
 import androidx.core.view.doOnDetach
+import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewTreeLifecycleOwner
+import com.twilio.livevideo.app.R
+import com.twilio.livevideo.app.custom.BaseLifeCycleComponent
 import com.twilio.livevideo.app.databinding.ParticipantViewItemBinding
 import com.twilio.livevideo.app.manager.room.ParticipantStream
 import timber.log.Timber
+import javax.inject.Inject
 
-class GridManager {
+class GridManager @Inject constructor() : BaseLifeCycleComponent() {
+    private lateinit var gridLayout: GridLayout
     private val participants: MutableList<ParticipantStream> = mutableListOf()
     private val bindings: MutableList<ParticipantViewItemBinding> = mutableListOf()
 
-    fun addParticipant(gridLayout: GridLayout, participantList: List<ParticipantStream>) {
+    private val _onStateEvent: MutableLiveData<GridManagerEvent?> =
+        MutableLiveData<GridManagerEvent?>(null)
+    val onStateEvent: LiveData<GridManagerEvent?>
+        get() {
+            val event = _onStateEvent
+            _onStateEvent.value = null
+            return event
+        }
+
+    fun init(lifecycle: Lifecycle, gridLayout: GridLayout) {
+        super.init(lifecycle)
+        this.gridLayout = gridLayout
+    }
+
+    override fun onDestroy(owner: LifecycleOwner) {
+        super.onDestroy(owner)
+        Timber.i("onDestroyCallback")
+        clean()
+    }
+
+    fun addParticipant(participantList: List<ParticipantStream>) {
         if (participants.isEmpty() && participantList.isEmpty()) return
         val updatedParticipantListUI = participantList.take(GRID_MAX_PARTICIPANTS_LIMIT)
         val participantListUI = participants.take(GRID_MAX_PARTICIPANTS_LIMIT)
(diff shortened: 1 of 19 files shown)


apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/RoomManager.kt
```diff
@@ -126,18 +123,28 @@ class RoomManager @Inject constructor(
...
+
         twilioException?.code?.apply {
-            if (this == TwilioException.ROOM_ROOM_COMPLETED_EXCEPTION) {
-                disconnect(RoomDisconnectionType.StreamEndedByHost)
+            when (this) {
+                TwilioException.ROOM_ROOM_COMPLETED_EXCEPTION -> disconnect(RoomDisconnectionType.StreamEndedByHost)
+                TwilioException.PARTICIPANT_NOT_FOUND_EXCEPTION -> disconnect(RoomDisconnectionType.SpeakerMovedToViewersByHost)
             }
+        } ?: run {
+            disconnect(RoomDisconnectionType.SpeakerMovedToViewersByHost)
```
timrozum, line 140: What happens when there is some random error that we don't specifically handle? Don't we need to clean up and pass that error to the UI to display? I don't see how this code handles this case but maybe I don't understand the Kotlin.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/GridManager.kt
```diff
@@ -124,13 +158,32 @@ class GridManager {
             val inflater = LayoutInflater.from(gridLayout.context)
             val participantViewBinding: ParticipantViewItemBinding =
                 createParticipantView(inflater, item, gridLayout, isVerticalModeEnabled, lifecycleOwner, indexView)
+
+            item.identity?.apply {
+                val popup = createSpeakerPopup(this, participantViewBinding.participantMenu)
+                participantViewBinding.participantMenu.setOnClickListener {
+                    popup.show()
+                }
+            }
```
timrozum, line 167: Is there a good way to move this from the grid code to the participant view code? For these reasons:

1. The grid code would be more single responsibility. It would be responsible for adding, removing, or moving grid items but not the internal behavior of a grid item.
2. When the speaker layout is implemented this popup will need to be handled there also. If the popup code remains in the grid code, it would need to be duplicated for the speaker layout.

timrozum (changes requested):

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/RoomManager.kt
alejo740: Yeah, this is cleanin up. I if you go to the `disconnect` method implementation (Line 72)... you will find it, internally it's cleaning up:
```
private fun disconnect(disconnectionType: RoomDisconnectionType?, notifyStateEvent: Boolean = true) {
        cleanUp()
        if (notifyStateEvent)
            _onStateEvent.value = RoomViewEvent.OnDisconnected(disconnectionType)
 }
```

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/GridManager.kt
alejo740: I have chosen to implement the `Popup menu` here instead of Participant code, because this is related to UI scope. So, in the participant code, I don't keep any code about UI. 

I understand your reasons, so, In that sense, I would prefer to move that code to the Screen code (Aka Fragment), then it can be shared with the `Speaker Layout` in the future.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/RoomManager.kt
timrozum: How does this error handling cover other errors like `TVIErrorParticipantDuplicateIdentityError = 53205`?

Errors that do not require specific handling should still be handled in a generic way. Can you explain the path for this case?

For example this is how iOS handles other errors: https://github.com/twilio/twilio-live-interactive-video/blob/fea431032bffde35f5fb93b6ff8afb2caf95c3de/apps/ios/LiveVideo/LiveVideo/Twilio/Room/RoomManager.swift#L91

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/GridManager.kt
timrozum: Ok, I will leave it up to you if you want to refactor now or later when you implement the speaker layout.

alejo740: I was checking the impact of these changes. I would prefer to make those changes once I start working on the `Speaker Layout`  implementation.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/room/RoomManager.kt
alejo740: done!

PR #163, opened 2022-09-09
Task/video 11148/android sync sdk integration raise hand feature

alejo740:
### JIRA link(s):
- [VIDEO-11148](https://issues.corp.twilio.com/browse/VIDEO-11148)
- [VIDEO-11002](https://issues.corp.twilio.com/browse/VIDEO-11002)

### Description

- Implemented SyncSDk along with UserDocument, RaiseHandMap, ViewersMap and SpeakersMap
- Implemented Raise-hand feature for the Viewer
- Implemented API call raise-hand to notify the user has raised the hand.
- Implemented API call send-speaker-invite but it's not being used yet.

### Demo
https://user-images.githubusercontent.com/929504/189435755-80203439-0bd6-4aa6-b837-e958bdedada2.mov


apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/sync/SyncDocumentWrapper.kt
```diff
@@ -0,0 +1,67 @@
...
+            _onStateEvent.value = null
+            return event
+        }
+
+    private val syncDocumentObserver: SyncDocumentObserver = object : SyncDocumentObserver() {
+        override fun onUpdated(context: EventContext?, data: JSONObject?, previousData: JSONObject?) {
+            super.onUpdated(context, data, previousData)
+
+            data?.apply {
+                val inviteKey = "speaker_invite"
```
timrozum, line 33: This code makes this object specific to the `user` sync document. And so I think it would be good to have `user` somewhere in the name of the class. Otherwise the name of this class implies it is for any kind of sync document.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/sync/SyncMapWrapper.kt
```diff
@@ -0,0 +1,102 @@
...
+import com.twilio.sync.EventContext
+import com.twilio.sync.SuccessListener
+import com.twilio.sync.SyncClient
+import com.twilio.sync.SyncMap
+import com.twilio.sync.SyncMapObserver
+import com.twilio.sync.SyncMapPaginator
+import com.twilio.sync.SyncOptions
+import org.json.JSONObject
+
+class SyncMapWrapper constructor(private var uniqueName: String = "") : SyncObject {
```
timrozum, line 15: The name of this is less of an issue than the document class, but I think the same logic applies. The name currently implies that this class is to be used with any sync object that is a map, which isn't really true. This class is really only to be used for sync maps that have this specific structure where the identifier is the user identity. We could add other maps in the future that are used completely differently.

This is minor stuff, but I think useful to have a less generic name.

timrozum (approved):
Approved but had a coupling naming suggestions.

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/base/LiveVideoApplication.kt
```diff
@@ -15,6 +16,7 @@ class LiveVideoApplication : Application() {
         super.onCreate()
 
         if (BuildConfig.DEBUG) {
+            SyncClient.setLogLevel(SyncClient.LogLevel.VERBOSE)
```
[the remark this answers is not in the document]
alejo740: Yeah, that's a good idea of having it with a lower log level, but in case of the SyncSDK, the default level is `SILENT` which means `Show no log messages.`. So, according to the log levels in Android, I changed to `DEBUG` to reduce the logs. 

FYI, These are the log levels we have in android SDKs:
```
        /** Show low-level tracing messages as well as all Debug log messages. */
        public static final int VERBOSE = android.util.Log.VERBOSE;

        /** Show low-level debugging messages as well as all Info log messages. */
        public static final int DEBUG = android.util.Log.DEBUG;

        /** Show informational messages as well as all Warning log messages. */
        public static final int INFO = android.util.Log.INFO;

        /** Show warnings as well as all Critical log messages. */
        public static final int WARN = android.util.Log.WARN;

        /** Show critical log messages as well as all Fatal log messages. */
        public static final int ERROR = android.util.Log.ERROR;

        /** Show fatal errors only. */
        public static final int ASSERT = android.util.Log.ASSERT;

        /** Show no log messages. */
        public static final int SILENT = Logger.SILENT;
```

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/sync/SyncDocumentWrapper.kt
alejo740: Done!

apps/android/LiveVideoApp/app/src/main/java/com/twilio/livevideo/app/manager/sync/SyncMapWrapper.kt
alejo740: Done!

afalls-twilio (approved):

PR #164, opened 2022-09-13
Task/video 11149/android allow play audio on speaker

alejo740:
### JIRA link(s):

- [VIDEO-11149](https://issues.corp.twilio.com/browse/VIDEO-11149)

### Description

- Enabled streaming to play audio on speaker instead of earpiece


afalls-twilio (approved):
As long as we don't need to support other outputs.

